Issue #184, opened 2022-07-13
Add missing build dependencies to Jenkins CI


jorge-campo:
Supersedes: https://github.com/status-im/help.status.im/issues/157

**Problem:**
The Material for MkDocs site generator requires a few `pip` and non-`pip` (`apt`/`brew`) packages in addition to the MkDocs ones. So far, we have only included the pip packages in `requirements.txt`:

```
pyyaml==6.0
mkdocs==1.3.0
mkdocs-material-extensions==1.0.3
mkdocs-git-authors-plugin==0.6.4
mkdocs-git-committers-plugin-2==0.4.3
mkdocs-git-revision-date-plugin==0.3.2
mkdocs-git-revision-date-localized-plugin==1.0.1
git+https://${GH_TOKEN}@github.com/squidfunk/mkdocs-material-insiders.git
```

**Solution:**
Include the non-`pip` packages (`apt` packages, for Linux) in the CI scripts.

**Non-pip packages required for Status Help:**
`apt-get install libcairo2-dev libfreetype6-dev libffi-dev libjpeg-dev libpng-dev libz-dev`

(cc: @exebeeche)

jorge-campo:
Requirements documented for contributors in `CONTRIBUTING.md` via https://github.com/status-im/help.status.im/issues/164.

PR #189, opened 2022-07-13
ci: use Dockerfile for Jenkins CI builds

jakubgs:
This way we have control over our system dependencies.

Resolves: https://github.com/status-im/help.status.im/issues/184

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,15 +1,19 @@
 pipeline {
   agent {
-    label 'linux'
+    /* Uses Dockerfile at repo root. */
+    dockerfile { label "linux" }
   }
 
   options {
-    disableConcurrentBuilds()
+    timestamps()
+    /* Prevent Jenkins jobs from running forever */
+    timeout(time: 10, unit: 'MINUTES')
     /* manage how many builds we keep */
     buildDiscarder(logRotator(
       numToKeepStr: '20',
       daysToKeepStr: '30',
     ))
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -38,11 +42,9 @@ pipeline {
 
     stage('Build') {
       steps {
-        sh "python -m mkdocs build -f config/en/mkdocs.yml"
-        sh 'python -m mkdocs build -f config/style-guide/mkdocs.yml'
+        sh 'python3 -m mkdocs build -f config/en/mkdocs.yml'
         /* Temporary solution for lack of start page. */
         sh 'cp overrides/static/* generated/'
-        echo "${GIT_BRANCH}"
```
jorge-campo, line 45: Hey @jakubgs, where is now this `GIT_BRANCH` parameter?

jakubgs: >where is now this GIT_BRANCH parameter?

What do you mean? That was just a useless `echo` leftover from some debugging.

jakubgs: Also, check now, it was cached.

jorge-campo:
Superseded by https://github.com/status-im/help.status.im/issues/195.

jakubgs:
What do you mean by "superseded"? "Related", sure, but it would be "superseded" if that PR had a different and newer implementation of the build setup, not a fix.

PR #250, opened 2022-10-06
Fixing the perfect-forward-secrecy URL typo

shamisem:
The typo is fixed in both the includes/urls-en.txt file and 'About Status messages', the article that mentions forward secrecy.

### DIFF
--- a/docs/en/using-status/about-status-messages.md
+++ b/docs/en/using-status/about-status-messages.md
@@ -48,7 +48,7 @@ The Status app secures your messages the entire time they are in transit using e
 
 When you communicate with someone on the Status app, your messages, attachments, sender metadata, group chats, and group metadata are all end-to-end encrypted.
 
-Status incorporates the [Perfect Forward Secrecy :octicons-tab-external-16:][perfect-foward-secrecy]{:target="_blank"} (PFS) encryption mechanism, ensuring that encryption keys change on every message. If your keys are compromised, only the associated message is compromised. All previous messages remain private.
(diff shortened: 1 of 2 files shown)


jorge-campo (approved):
Thanks for the fix, @shamisem!

Could you please now merge the fix into `develop` and then into `master`? (let me know if you get stuck at any point.)

- For merging into `develop` look [here](https://www.notion.so/Technical-writer-s-reference-guide-0b3bef406f914bce95d6b340c6d427e9#7a09cbdfb10e4b5a9809a714710795b0).
- For merging into `master` look [here](https://www.notion.so/Technical-writer-s-reference-guide-0b3bef406f914bce95d6b340c6d427e9#775409b21b0c40b1a136fc06d7dd52d7) (you don't need to create a new release.)

Please, remember:

- Every task (whether a fix, a new doc, or any non-doc tasks) has [an associated issue](https://www.notion.so/Technical-writer-s-reference-guide-0b3bef406f914bce95d6b340c6d427e9#dbb21d31cde34ce888208e969c861b5f) in the [GitHub Project board](https://github.com/orgs/status-im/projects/71/) (for a small thing like this we don't need an issue! :) but let's create one in this case, for the exercise)
- Your feature branch (`fix-forward-secrecy-url-typo`) should have the issue ID as a prefix. So, if the issue ID is 267 (for example), the branch name is `267-fix-forward-secrecy-url-typo`. You can ignore this rule in this fix and go ahead with the merge.
- We don't have rules to name commits (`git commit -m` switch), but I find it helpful to add the same issue ID as a prefix.

PR #253, opened 2022-10-14
Edit and delete your messages

shamisem:
Procedural topic explaining how to edit and delete messages in Status Mobile.

Related to #249.


jorge-campo (changes requested):
Thanks for this article, @shamisem — I write here the things to change:

- [x] Set cross-links to other content, like when you mention the peer-to-peer network concept. There is an article explaining what's a peer-to-peer network (see [here](https://help.status.im/en/using-status/about-status-messages/)). Explaining these concepts is especially important in our case because users coming from other platforms are not familiar with them.
- [x] When introducing a new concept, cross-link an existing article or, if necessary, a link to an external source (like Wikipedia). If there is no article in Status Help for that concept, and you think there should be, add the article to the Miro mind map.
- [ ] You don't mention how long I can edit or delete my messages. This is important.
- [ ] ~~Why if I want to edit/delete a previous message, but it's so old (>30 days) I can't find them on the conversation timeline? You may need to follow up on the conversation with Jo to find out (see [Discord thread](https://discord.com/channels/624205794384281629/697843216048717905/1010277726822875156)).~~
- [x] We use the tabbed interface even when we only discuss one platform (in your case, mobile). Please, see the Material for MkDocs documentation for the tab-layout syntax.
- [x] Every procedure starts with where the action occurs (see [Writing tasks](https://help.status.im/style-guide/structuring-the-content/#writing-tasks)). In your case, that place is the message list.
- [x] "you'll need other members' Status apps to process and apply these changes." This is confusing. It makes me think I should do something as a user, but there is nothing I can do to avoid this situation. We must explain that message changes (like edit or delete) propagate by default to all participant nodes in a peer-to-peer network. Still, if a node is disconnected or running a modified (or out-of-date) Status app version, the message may appear on another user's device. In summary, there is nothing the user can do to avoid this, but your sentence will make the user thinks it is.
- [ ] Does this procedure apply to all types of messages? Can I delete a picture I sent by mistake? Ask developers or designers as necessary.
- [ ] Can I delete messages for every type of message (1:1, group chat, public chat, and communities)? I guess so, but better to double-check.
- [x] Your admonition goes in the middle of the intro, which isn't allowed in the procedure structure. Admonitions in intros go right before the procedure (see [Procedure structure](https://help.status.im/style-guide/structuring-the-content/#procedure-structure))
- [ ] You added the icons for the icon on a different PR. Try to have all the commits related to an article in the same PR so that when I fetch the PR, I can fetch all the data.
- [x] Have you previewed this article in your local Material for MkDocs build? The admonition header is missing because you use the double-quotes pair (`""`). This admonition should include the header.
- [x] "Be careful with _the_ information you share" (?)
- [x] Add the article's name in the using-status/index.md file

Update 18.Oct: I deleted one entry because I realized it didn't make sense. If I have written a message, it is always on my device, so I always have the option to get back to that message and delete it.

jorge-campo:
I am working on the review.

jorge-campo:
Superseded by https://github.com/status-im/help.status.im/pull/283.

PR #293, opened 2022-11-19
Create a group chat

shamisem:
Procedural topic explaining how to create group chats in Status Mobile and Desktop.

Note: I had to cherry-pick a commit to prevent potential merge conflicts


jorge-campo (changes requested):
Great article, @shamisem ! Thanks for this.

Here are some notes:

- Spell out numerals < 10. See https://write.status.im/style-conventions/?h=num#numbers-date-and-time-currencies-and-units-of-measure.
- "To invite someone to a group chat, you need to have them on your contact list." Great that you mention this; it's important. But be careful here; we don't want users to confuse a Status contact with a contact they have in their phone book. I would say something like this "To invite someone to a group chat, you need to have them on your Status contact list."
- When you add an internal link, create the new empty Markdown file the link is pointing to. For example, for the [create-a-status-community] link, create the `/docs/en/status-communities/create-a-status-community.md` file and add the following text right below the "Create a Status community" H1 title (we'll discuss this in more detail in a future Deep Dive session; see [this Notion page](https://www.notion.so/Identifying-articles-by-development-or-publication-stage-3e35a44a4d464e10882729d127e8dd2a)):
	```
	:octicons-tools-24:
	
	!!! note ""
	    We're working on this content.
	```

- We can improve the article by mentioning+linking other things users may want to do on a group chat. Alternatively, you can write this in the "Understand group chats article":
	- Mentioning (@) someone else – This is useful in group chats that grow in number or very active groups, where it's easy to miss something directed to you.
	- Pin messages – Please, double-check this. Can we pin messages? Who can pin them, anyone or the group owner?
- Remember: ["In the first step, tell the user where to start"](https://write.status.im/structuring-the-content/#writing-tasks). The (+) icon you mention doesn't appear on every screen. The procedure should start more or less as follows:
	- For mobile: "From the tab bar, tap **Messages** [Messages icon]" (You will need to add the "Messages" mobile icon.)
	- For desktop: "From the navigation sidebar, click **Chat** [chat icon]"

(Note: ideally, these actions should be consistent, so I have opened an issue with the Desktop team to rename "Chat" to "Messages;" see https://github.com/status-im/status-desktop/issues/8418)

- "Optionally, type a message to start the group chat with in the Message text field." (Desktop procedure, step 4). This is perfectly valid but hard to read for non-natives. Besides, remember that location goes before action (see https://write.status.im/structuring-the-content/?h=act#writing-tasks). I would write something like this "Optionally, type a message in the Message text field to start the group chat."
- We want to emphasize what makes Status different and better than other solutions. We're not (yet) as feature-full as Messenger or Telegram, but all Status communications are encrypted by default. In the "What to expect" section, you can add something like this: "Group chats are always end-to-end encrypted with strong cryptographic keys not known to Status. Only the group participants can read the chat messages." You can also link to the "About Status messages" article > "Built-in privacy and security" H2.

jorge-campo (approved):
Thanks for the updates, @shamisem !

shamisem:
Merged!

PR #318, opened 2022-12-15
Remove a contact from Status

jorge-campo:
Document for your review, @shamisem!

Thanks.


shamisem (commented):
Hey Jorge! Here's my review:

- "In Status, you must add people as contacts to start a conversation"
"<...> you need to add people as contacts <...>" sounds more natural to me in this case.
- "After you remove a Status contact, you can't start a conversation, and your contact can't send you new messages."
It sort of sounds as if these are two different restrictions. Maybe we can do something along the lines of "After you remove a Status contact, neither you nor your contact can send new messages to each other" instead?
- "Removing a contact from Status removes the identity verification associated with that contact but not its reputation (trusted or untrusted)."
I'd say "their reputation" since we're talking about a person.
- "After removing a contact, your direct messages with the contact remain"
I think it's better to phrase this as "After you remove a contact, <...>".
To avoid repetition, maybe we should also replace the second "contact" with "them" ("your direct messages with them remain").
I'm also not sure we should be talking about removing messages as opposed to deleting them, since as far as I know messages can't be recovered in Status once you [remove/delete] them. The second word implies exactly that, so perhaps that would be the better term?

jorge-campo:
Great review, @shamisem ! Thank you.

I'm applying the changes and pushing them onto develop and master.

About "must"/"need to":

Our writing style guide [advises against using "need to."](https://write.status.im/style-conventions/#common-words-usage) It may sound more natural, but it's also less assertive. I added this intentionally when I wrote the style guide, based on my experience with other style guides.

I have added a note to talk about this in our next meeting and, if necessary, update the style guide.

jorge-campo:
Superseded by https://github.com/status-im/help.status.im/pull/341.

PR #319, opened 2022-12-19
Add members to a group chat

shamisem:


jorge-campo (changes requested):
The article is correct, @shamisem, but misses some extra information to make it more helpful. When you write an article, no matter how simple it may look, put yourself in the reader's seat. Anticipate their questions or where else they may want to go.

### Things to change

- I prefer "... can add their Status contact to *the chat*" and omit the preposition. It sounds less natural and more repetitive, but I keep readers with limited English in mind.
- Link "[...] Status contacts [...]" in the first sentence to the "Send a contact request" article (doesn't exist yet; change its title if necessary.)
- If you point the user to a different article in Status Help, the action verb is "check," not "see." This is explained in our writing style guide (see [User actions](https://write.status.im/style-conventions/?h=check#user-actions)).
- In the mobile how-to, "this will say **Add / remove members**" is missing the icon.
- In the desktop how-to, you're missing where to start. The first step should read: "1. From the navigation sidebar, Open the ..."
- The article's entry in the `mkdocs.yml` `nav` section is missing. The article won't appear in the hamburger menu.
- The article's URL in the `urls-en.txt` file is missing.

### Information I miss

- What's the maximum number of members in a group chat? It's worth mentioning this information in a tip admonition, or as part of the first paragraph.
- After the members limit is reached, and a user wants to add another member, what occurs on the screen? The Figma files mention this option, but I can't see the resultant user interaction ([link to Figma file](https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=668%3A225911&t=I9LyJQys4g9oedjY-0).) You can ask the designers first and, if they don't know, the developers in the Status CC Community.

Screenshot 2023-01-18 at 4 39 14@2x

- Ideally, when we explain the how-to, we can indicate something like, "If the **Add members** option is grayed out, the group chat has reached the maximum number of members, and you can't add more." (this is just an example, I don't know if the option is grayed out.)
- If I join a group chat, can I see the previous chat history? I assume I can see at least 30 days back (the maximum number of days Waku stores this info), but I'm not entirely sure.

Please, update the article accordingly. If you can't get the article updated by Friday, discard your current local branch and work on a new one without pushing the content, so we can start with the icons cleaning process.

Thanks.

PR #321, opened 2022-12-20
Delete a group chat

shamisem:


jorge-campo (changes requested):
Thanks, @shamisem -- Here are my review comments.

- Let's try to reduce the following elements as much as possible:

	- passive forms ("can be deleted")
	- pronouns ("it")
	- future forms ("will be gone")

	I prefer to use 'you' repeatedly and replace pronouns with nouns, even if it may sound monotonous. Use Wordtune to find alternatives.

- Move "Only the group chat owner can delete it. If you're a regular member, consider [leaving the chat][leave-a-group-chat] instead." into a note-style admonition.

```
You can delete a group chat if you don't need the group anymore. The chat history, member list and all the other data are gone forever when you delete the group chat.

!!! note
	Only the group chat owner can delete the group. If you're a regular member, consider [leaving the chat][leave-a-group-chat] instead.
```

- I would add a cross-link to 'Understand group chats' in < ... you can delete a [group chat][_cross-link_] if you ...>
- You're missing the icons in the procedural steps. Remember: when referencing an entry on a list or UI control with an icon, include the icon before the name. This rule is not only valid for buttons, but for any other UI element the user can interact with.
-   Start procedures telling the user where, even if this sounds easy enough.
    
    For mobile: 1. From the tab bar, tap [messages-icon] **Messages**.  
    For desktop: 1. From the navigation bar, click [chat-icon] **Chat**.

- As a general comment, bear in mind and apply our style guide rules.

PR #326, opened 2022-12-22
Join a group chat

shamisem:


jorge-campo (changes requested):
Short and to the point, as we discussed.

I would add a cross-link to 'Understand group chats' in < ... the only way to join a [group chat][_cross-link_] is to ask ...>   

 You're still missing:
 
 - `/using-status/index.md` entry for the article.
 - `nav` entry for the article in `mkdocs.yml`.

Please, fix these and merge your changes.

Thanks!

Issue #287, opened 2022-11-17
Channel quick start guide


Fabiomorais87:

jorge-campo:
For reference only -- This is an early version of this same document:

---
id: 111
revision: 0
language: en
hide:
- navigation
---

# What's a Community channel?

:octicons-cpu-24: In progress (1)
{ .annotate }

1. We’re working on this software feature. This information may be 
   incomplete or not work as expected.

A channel is a dedicated space where people talk, share ideas,
have discussions, and post content.
Status Communities organize people's communication into channels.

Community channels usually unite a common topic for conversations and 
discussions.
Using channels, Communities keep conversations engaged on a topic.

### The basics



- Users have access to all channels inside the Community after joining it.
- Users can't delete channels, but they can mute unwanted channels.
- Community owners can [create channels][create-a-community-channel] as well 
  as categories for their needs.

!!! tip

    Community users can mute channels, mark messages as read, and delete 
    message history in channels. See more info on the 
    [Use channel actions][use-channel-actions] page.



## The #general channel

This channel is typically used for announcements and general discussions. 
Also, users can get familiar with Community FAQ and rules.

A new Community starts with the #general channel by default.
If you are a Community owner, you can change the #general channel or delete 
it if you don't need it.


## Categories

A category is like a folder where Community owners can place channels 
according to their topic.
Using categories, users can navigate faster in a Community.

When you run a Community, you can arrange channels into categories for 
convenience.
Use drag and drop actions to build an effective architecture for Community
channels.
These actions work for channels as well as for categories.



--8<-- "includes/urls-en.txt"

PR #334, opened 2023-01-06
287 Channel your quick start guide

Fabiomorais87:
Hi @jorge-campo ,

Here is the PR for the article Channel your quick start guide. I'd like to hear your comments. 

The notion page "Screenshot requests" has been updated with the images requests.

Disclaimer: Do you remember that problem I mentioned to you and Emma with the icons? I experienced the same problem with this article. Let me know if it works when you open it. It doesn't show when I preview or build my site locally on my computer.

GitHub issue: https://github.com/status-im/help.status.im/issues/287


jorge-campo (changes requested):
Thanks for your work, @Fabiomorais87.

We need to review this article together. After looking into the Figma workflow again, I have the impression we're only scratching the surface here.

In summary, you can have access to channels in read only mode even when you're not a community member. This is a bit confusing, and requires further explanation on our side.

I'm closing this PR to keep things clean in preparation to the icon cleaning process. Please, don't discard your local branch.

PR #335, opened 2023-01-07
295 Leave a channel

Fabiomorais87:
Hi @jorge-campo ,

Here is the PR for the article Leave a channel. I'd like to hear your comments.

On **"you leave the community"** I linked to an article called "how-to-use-communities-your-quick-start-guide.md" My first thought was to link to an article named "leave a community" but it did not exist. In the Miro map, I found this link option (quick start guide). Please let me know if creating a place holder for the article "leave a channel" would be more appropriate. In that case, we will need to update the miro map as well as create an issue on Github.

### DIFF
--- a/config/en/mkdocs.yml
+++ b/config/en/mkdocs.yml
@@ -166,6 +166,8 @@ nav:
     - status-communities/index.md
     - Status Communities:
       - Join a Status Community: status-communities/join-a-status-community.md
+    - Channels:
+      - Leave a channel: status-communities/leave-a-channel.md
   - Wallet:
     - status-wallet/index.md
   - Your profile:
--- a/docs/en/status-communities/about-losing-access-to-a-community.md
+++ b/docs/en/status-communities/about-losing-access-to-a-community.md
@@ -1,0 +1,15 @@
+---
+id: 301
+revision: 0
+language: en
+hide:
+  - navigation
+---
+
+# About losing access to a community 
+
+!!! note ""
+     We're working on this content.
+
+
+--8<-- "includes/urls-en.txt"
--- a/docs/en/status-communities/channels-your-quick-start-guide.md
+++ b/docs/en/status-communities/channels-your-quick-start-guide.md
@@ -1,0 +1,15 @@
+---
+id: 287
+revision: 0
+language: en
+hide:
+  - navigation
+---
+
+# Channels: your quick start guide
+
+!!! note ""
+     We're working on this content.
+
+
+--8<-- "includes/urls-en.txt"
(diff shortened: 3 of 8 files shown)


jorge-campo (changes requested):
## Rethinking the topic

After thinking about this article, I'm shaking and changing things. Let me know what you think.

- We have an article called "About losing access to a community," and I think something like this for channels makes more sense than "Leave a channel."
- Allow me to explain. Our initial assumption that readers will look for information about how to leave a channel is wrong. Users are already familiar with the fact that they have access to channels when they join a community. What they're not aware of is that they can lose access to them, and this is what we should explain.
- This also helps us connect with the content in "About losing access to a community."
- Because of this, I changed the article's name to "About losing access to a channel. "

Note: Don't get stuck on what the Miro mind map says; adapt and change things as necessary.

## Review notes

Please, take them into account for future articles.

- Don't use Markdown links (`[content](link)`) except for linking to a sub-headers. Use links as described in our writing style guide. Check [Manage and format URLs](https://www.notion.so/Writing-content-guide-41df56af6140402894819ccdf28fa0f7#863818fb7a7549dbb7f2b2388dda1a31) in our writing guide.

- You're missing the bold typeface in the article's name at `docs/status-communities/index.md`.

From our style guide:
> At the index page level, use a bold typeface to stress the action and the feature or concept in the document.

- The article's entry in the `urls-en.txt` file should be sorted alphabetically. Your entry wasn't in alphabetical order. This is important because we'll typically have merge conflicts in this file, and keeping things tidy help to fix the conflicts faster. This is what you wrote:

```
[create-a-status-community]: /en/status-communities/create-a-status-community
[join-a-status-community]: /en/status-communities/join-a-status-community
[what-is-a-community-channel]: /en/status-communities/what-is-a-community-channel
[leave-a-channel]: /en/status-communities/leave-a-channel
```

- Remember one paragraph, one idea. Don't mix too many things in the same paragraph. This was a problem in your initial paragraph.

- I've rewriting some sentences based on the content we have in https://github.com/status-im/help.status.im/pull/333/files#diff-91516b13e130665f2804ce6c208c1ba2f5bf679b89141c18421494b0f74f275eR15-R32. Reuse content as much as you can; it makes the content easier to read and understand across different articles.

- You mention "open" and "token-gated" channels. The user may ask how to distinguish between them in the UI. A callout there can clarify this quickly. I'm reusing the same text from https://github.com/status-im/help.status.im/pull/333/files#diff-91516b13e130665f2804ce6c208c1ba2f5bf679b89141c18421494b0f74f275eR15-R32.

- The items you list in the bullet points are missing a final period. See [Bullet lists](https://write.status.im/style-conventions/#bullet-lists) in our writing style guide.

jorge-campo:
Superseded by https://github.com/status-im/help.status.im/pull/359.

PR #336, opened 2023-01-09
Clear group chat history

shamisem:


jorge-campo (dismissed):
Thank you, @shamisem !

Here are my review comments:

- Don't forget to add the "Project" metadata to your PR after sending it, and to assign the "Review" status (right-hand side in the GitHub UI.)

- Please, update the Miro mind map entry with the issue ID and GitHub link.

- You're missing the article's entry in the `nav` section in the `mkdocs.yml` config file, so you're article won't show up in the hamburger menu. Let me know if you need help to add this entry.

Screenshot 2023-01-10 at 11 41 32@2x

- The index entry in `urls-en.txt` is not correct and the link in `index.html` fails.
	- `[clear-group-chat-history]: /en/clear-group-chat-history` should be `[clear-group-chat-history]: /en/using-status/clear-group-chat-history`

- We have a problem where some icons, for reasons we still need to determine, don't show-up. Icons in your document don't up for me, but I can't see the `mobile-more` and `desktop-more` icons in your `320-clear-group-chat-history` git branch either. Could you please add the icons to your `320-clear-group-chat-history` branch?

Screenshot 2023-01-10 at 11 10 06@2x

Screenshot 2023-01-10 at 11 09 55@2x


- Remember to start procedures telling the user where to start, even if this sounds easy enough.

	For mobile: 1. From the tab bar, tap [messages-icon] **Messages**.
	For desktop: 1. From the navigation bar, click [chat-icon] **Chat**.
	
- In the mobile instruction, step 3, you're missing the "trash" icon. If the option we refer to uses an icon, we also write the icon in the instruction: Select [mobile-trash] **Clear history**. (you'll need to add the "trash" icon to your local branch.)

- The same goes for the desktop instruction; you're missing the desktop "trash" icon. Please, add the icon in the corresponding folder in your local branch.

- I would include an introductory sentence to clarify what we're talking about here. This may sound too obvious, but makes things easier for the reader:

	>Clearing a chat clears the messages inside the chat.
	
	Look for examples on the internet ([How to clear chats](https://faq.whatsapp.com/3267207290200464/?cms_platform=android) in the WhatsApp Help Center.)

- About the first paragraph:

	>Only the group chat owner can clear its history for all members. If you’re not the owner, you’ll clear it just for yourself. Everyone else will still see the full chat history.

	- The three sentences in the same paragraph make me doubt if the "Everyone else ..." sentence refers to the group chat owner or a regular group chat member. 
	- I would split the sentences into two different paragraphs and avoid the first future tense:

> Only the group chat owner can clear its history for all members.

>  If you’re not the owner, you clear the chat history just for yourself. Everyone else will still see the full chat history.

PR #358, opened 2023-01-18
Update download status for mac

cheny0:


jorge-campo (changes requested):
Thanks for the update, @cheny0.

Just a minor thing: remove the following sentence: "Take a look at the steps below to get started." We don't use these meta information where we tell the reader what s/he is going to read. This was in the original document, but it's an old one and we didn't have the style guidelines yet.

When you send a PR for review, set the issue status to "review". You can do this action from the PR right-hand side bar:

Screenshot 2023-01-18 at 5 03 14@2x

To complete this review you'll need to:

- Remove the sentence and push your changes again.
- Merge your changes into `develop`. There is no need for an additional PR+review because the review phase is done. 
- Verify in Jenkins and in https://dev-help.status.im that the updated article is available.
- From `develop`, send a PR with the updated content to `master`.
- Merge the PR into `master`

> **Warning**
> Do not use the GitHub web to merge into `master`, use the CLI

- Again, verify in Jenkins and in https://help.status.im that the updated article is available.
- Update the GitHub project board with any issue or PR that needs to be updated.

PR #385, opened 2023-01-27
Pin a message

### LINKED ISSUE #339: Pin a message

earakel:


jorge-campo (changes requested):
Thank you, Emma.

The document is correct, but I miss some additional information to make it not only accurate but useful. Why not include information about why it is helpful to pin a message? Why not give users ideas on how to use this functionality? Users probably understand what it means to pin a message, but they haven't considered good use cases. For example:

- Keep important information accessible -- Like the rules to follow on a particular channel.
- Highlight announcements -- For example, when a new product or software version is available.
- To track progress on a project -- Updating the pinned message after a project milestone.
- Save frequently used information -- Like where to find a frequently-used document or what command to run to something.
- Create a to-do list -- For example, when you plan a trip with your friends on a group chat.

These are just examples. Adding some (or all of them) to the article would make it more relevant.

**Other review comments**

- You're missing the article's entry in the `/using-status/index.md` file, so nobody will see the article browsing Status Help. 
- Add the article's entry to the `mkdocs.yml `so users can find it from the hamburger menu.
- The first sentence is 100% perfect English, but it's hard to read for non-native speakers. What do you think of something like "Chat members can pin both their own and other members' messages"?
- "DM" is widely used to refer to 1-to-1 chats, and we use it already in other docs. I suggest:
	1. Add DM in the first paragraph: "Message pinning is available in direct messages (DMs), group chats ..."
	2. Use DM in the last paragraph: "To view all pinned messages in a group or DM chat, open ..."
- The last paragraph needs a screenshot; only with the text, it's hard to figure out where to tap. Please add this one to our list in [Notion](https://www.notion.so/Simplified-screenshots-in-Status-Help-9b99304babb042e2a63f4bc94aa0b179).

shamisem:
Thanks for your review @jorge-campo! 

I've just updated the article, requested a screenshot, and created [an issue (#412)](https://github.com/status-im/help.status.im/issues/412) to update this article with the screenshot once it's ready. Re-requesting a review to make sure the article's good to go now

PR #419, opened 2023-02-06
Add a contact in Status

shamisem:


jorge-campo (changes requested):
Thanks, @shamisem.

- According to the Miro task analysis, there is a missing entry for this article in `direct-messages/index.md`. Either you add the entry or remove the Miro link.
- You're using a bold typeface to describe a user action inside and admonition. The style guide [mentions](https://write.status.im/style-conventions/?h=bold#format-guidelines):

>Use a bold typeface to write UI controls in procedural steps (1, 2, 3, and so on). Don't use the bold typeface outside of the procedural steps.

However, this is a valid use, and I have seen this in other docs ([example](https://slack.com/help/articles/218876627-Trouble-receiving-emails-from-Slack)). We should update the style guide; can you please update this? It can be something like:

>Use a bold typeface to write UI controls in procedural steps (1, 2, 3, and so on). Don't use the bold typeface outside the procedural steps except when asking the user to interact with the UI.
> Example:
>| Usage | Example |
>|:---|:---|
>| Correct | To add someone you know, simply right-click their name and select  **Send a contact request**. |
>| Incorrect | You can **Send a contact request** to add someone you know |

**Procedure**
- The article is missing adding a contact using a QR code; it only mentions the chat key link. [Here is the link](https://www.figma.com/file/PiN0uwRgZj4L9U4N7xO7Xb/Profile-for-Mobile?node-id=3773%3A221806&t=W1MRU8isbaTk247i-1) to the Figma file where this is discussed. I include an example below.
- Add a final tip, informing the user where to find the QR code for their profile (I'm aware this is discussed in the article you link to, but it's helpful in this context, nevertheless).
- ... which is a representation of their Status profile -> ... which is represents their _public_ Status profile.
- You can link to this content when you mention *chat keys*: understand-your-status-keys-and-recovery-phrase (this article is [under review](https://github.com/status-im/help.status.im/pull/423)).
- The paragraph you have right before the steps apply to Mobile and Desktop (it's the very same paragraph). Add the paragraph right after H2 and before the tabbed section so you don't need to repeat the same information.
- ... tap Messages -> ... tap **Messages** (in the Mobile tab)
- Oops, these are way too many steps under Mobile. We have a limit of six-ish steps, but nine is a no-go, I'm afraid. You have two options:
	1. Split the steps into two different groups -- This is what you can see in may other documents, and we're using this trick as well (see [Desktop section here](http://127.0.0.1:8000/en/getting-started/run-the-status-app-for-the-first-time/#create-a-new-status-profile)).
	2. Combine steps (in this case, I would choose this one)

Here is an example (I omit formatting and existing icons):

1. Ask the person you want to add to your contacts to share their profile.
2. From the ta bar, tap Messages.
3. Tap Plus and then select Add contact.
4. Paste the link or tap [scan-icon] to scan the QR code.
5. Tap View profile > Send contact request
6. In your introductory ...
7. Tap Send contact request

!!! tip
To find your contact's QR code, tap Messages, and then, from the navigation bar, tap [qr-code-icon] QR code.

PR #422, opened 2023-02-06
React and reply to messages

shamisem:


jorge-campo (changes requested):
Thanks for the article, Emma.

- Are you sure about the user's React and Reply interactions? In the "Reply" section, you mention "3 Tap the message", but the Figma file shows "Long press" (see [here](https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments-for-Mobile?node-id=3338:349124&t=ynhRH9E3IDviD4aG-0))
- You mention a new element/concept, "Notification Centre", so you need to point the user to the corresponding doc. So far, the right one is Adjust your notifications > 346 Your Status notifications guide.
- You're missing the correct format in the procedural steps. Please, review the UI labels format.
- Open the chat with the message you want to react to -> Open the chat *or channel* with ... This applies to both procedures in both platforms.
- Reply to messages: Open the chat with the message you want to react to -> ... with the message you want to *reply* to.
- You can also reply with a picture or a sticker, not only by typing. A "tip" admonition with this information at the end of the procedure would be nice.
- For desktop, I'd write the two "hover over the message" and "click" steps in a single step.
- Please, remove the duplicate entries for the articles in the "Using Status" `index.md` file. As discussed, we want to have one entry only for each article.

Screenshot 2023-02-16 at 11 05 38@2x

PR #444, opened 2023-02-21
Share your Status profile

shamisem:


jorge-campo (changes requested):
Thanks, @shamisem ! The article is great, but I propose some additions to make it even better.

I'm using [this Figma file](https://www.figma.com/file/PiN0uwRgZj4L9U4N7xO7Xb/Profile-for-Mobile?node-id=3802%3A260430&t=9Y3fOQzoI75WLWXW-0) for reference.

1. In Mobile, step 2, you use the QR icon with the "Share" label. This is not the name we want to use for this icon. I would use "QR code".
2. The "Share" icon has an arrow pointing up. This icon is also on the profile screen. With this icon, you can share your profile using the operating system share sheet. This is another way of sharing a profile that we should mention.
3. For the emojihash, the screen allows you to copy this hash, but what happens then? Can I paste the emojihash in a message and send this to someone? Maybe not to someone (because that person is not my contact) but in a channel or group chat. It isn't obvious how I should use the emojihash, and we should help the user here. Maybe with a tip admonition (?)
4. This article points to your other article, "Add a contact in Status". After looking at this (again), I think you can move step 1 in "Add a contact in Status" outside of the procedure and add a link to this "Share your Status profile". In other words: "Ask the person you want to add to your contact to [share their profile][share-your-status-profile]."

By moving this first step out of the procedure, we can include a cross-link and give more context to users while reducing the number of steps in "Add a contact in Status" from 7 to 6. Besides, the first step is common for Mobile and Desktop, so it makes sense to have this outside the procedure and right after the "Add a contact" H2.

If something isn't clear, let me know and we clarify this on a call.

Thanks.

PR #448, opened 2023-02-21
Invite friends to Status

shamisem:


jorge-campo (changes requested):
Thanks, Emma!

A minor thing: you need to include the entry in the [index.md page](https://github.com/status-im/help.status.im/blob/e61979cbdfe105419365db56878303360602767e/docs/en/getting-started/index.md#L12). Other than that, it's good to go.

PR #452, opened 2023-02-24
Adding category descriptions

shamisem:


jorge-campo (approved):
Looks good! The format doesn't "pop," but there isn't much we can do with Material for MkDocs. You could try embedding the text into an untitled admonition (?)

You can compare the options in `mkdocs serve` and choose what you like the most, and feel free to merge the changes.

Thanks.

shamisem:
> 

I tried out quite a few options and none of them seemed to click perfectly, so we'll have to do what we can I guess

PR #487, opened 2023-02-28
Sync your profile across devices

cheny0:


cheny0:
Both "sign in" and "log in" appears in Figma files, so I used both in this article. I'll check which one the Design team decide to use before MVP.

jorge-campo:
> Both "sign in" and "log in" appears in Figma files, so I used both in this article. I'll check which one the Design team decide to use before MVP.

Use "sign in" :) -- This is the option in the Mobile Onboarding file ([see Figma](https://www.figma.com/file/o4qG1bnFyuyFOvHQVGgeFY/Onboarding-for-Mobile?node-id=3068%3A332555&t=hIZNKsHAuXyZjdMQ-0)), which is mature and ready for the MVP. Whenever you see "log-in", add a Figma comment explaining that the label should be "sign in".

jorge-campo (changes requested):
Thank you, Chen! 

Here are my review comments.

Your title mentions "profile" but then you mention "data" in different sentences. It's not entirely clear if these two things are the same thing. I would keep using "profile" consistently in your doc.

## Intro

- The first admonition, where you mention the feature availability, goes without a title (it's not a note, tip or warning). You can look at other examples for the Markdown code, like [Create a channel](https://help.status.im/en/status-communities/create-a-channel/).
- About the admonition: "Currently, you can only sync your profile using Status mobile." I think this is not what we want to say. While it's true that you can only use your phone to sync QR codes, it doesn't mean you can't keep Status on your smartphone in sync with Status on your desktop (correct me if I'm wrong, please). In other words, my desktop can participate in the sync process; therefore, it's in the scope of this topic even when I can't use my desktop to scan. Therefore, we need to clarify this further and probably remove the admonition (in a second paragraph?)
- [...] and you can switch devices seamlessly. -> I suggest using "[...] you can switch between them seamlessly" so we avoid repeating the word "device".

## About "What to expect"

- "Your messages are not synced, they are maintained using Waku" -> This is a limitation we need to confront openly with the user so they're informed of what to expect. What do you think about this? -> "Your messages are not part of your synced data. When you sync a new device, only the messages sent or received in the last 30 days are available on your new device. Messages you receive from that initial sync are always available."

## About the procedure

- I find the H2 titles missing the "profile" word. I understand it's implicit (on the article's title), but meaningful headers facilitate content search. Considering this, I'd clarify your H2s:

      - Sync your profile on a new device.
      - Sync your profile on a logged-in device.
      - Turn off profile syncing.

- You're missing the "Mobile" tab. Even if the procedure only applies to one of the platforms, we still use the tab for this platform to highlight from what platform to follow the procedure. See example [here](https://help.status.im/en/status-communities/create-a-channel/)
- Tip: [...] If the codes are expired, authenticate your profile again and get new codes". What do you think about this? -> "If the codes expire, authenticate your profile again and get new ones."
- The "Enable camera" icon is called "Camera".
- About the second step:
	- You use an icon without a label ("tap +"). Remember that all icons should go with a label in bold typeface and right after the icon. For this icon, the label is "Add"
	- Breadcrumbs are not meant to describe a sequence of events; they're meant to describe a hierarchy, which typically happens when you search for an option under your smartphone settings. The option you're looking for is deep-down in a specific options menu, and these menus follow a hierarchy and, in this case, a breadcrumb menu works. In your case, these options are simply on different areas of the UI, but they're not hierarchically related. Therefore, step 2 is "Go to Syncing" and step 3 is "Tap + Add and then tab Reveal sync code."
- The procedural steps don't follow the structure we're using in other docs (admittedly, this hasn't been updated in our Style Guide, so I see where the error is coming from.) In this case, you have a procedure with two parts and precede each part with an H3.
	- Your "In Status on your logged-in device" would be an H3 without the final colon.
	- Same thing for the "In Status on your new device".
- Same H3-based structure applies to the "Sync a logged-in device".
- In every procedure, the first step indicates where to start. You mention "Tap your profile photo", but what if I'm in the Wallet screen? There is no profile photo there (this is just an example). Your first step would be "1. From the tab bar, tap **Messages**" because we know that your profile picture is always in the messages screen when you tap there.
- About turn off syncing -- Unless the user (1) wants to dispose of one of their devices or (2) doesn't recognized a device on the list, we don't want them to turn off synching. I would add this information right before the steps. We want to prevent users from turning synching off because we know their experience using the app will degrade.

## Additional information

Do we need an article to explain how to authenticate users ("Authenticate your profile")? I have serious doubts. We discussed this the other day. The thing is, users know how to authenticate their profile; they know that they should use their fingerprint, face or password to access their information.

We may need to document how these elements work together. This is especially important for Keycard because if I use a Keycard, I don't know what happens during the authentication when using Keycard, along with other authentication methods. For example:

- If I use Keycard from the start, is it always defaulting to the Keycard prompt for authentication?
- If not, is it defaulting to my password?
- Can I use biometrics, password, and Keycard together to authenticate?

We must ask these questions in this hypothetical "Authenticate your profile" and not explain the steps to authenticate my profile. As I said, these are well-known by users.

In terms of terminology, I believe the expression "authenticate your profile" is a bit technical for our target audience. Something like "verify your identity" or "confirm your credentials" works better. I'm aware this would change not only some sentences in your article but also the title of the other article you're working on these days.

cheny0:
Change request and conflicts resolved.

PR #499, opened 2023-03-07
Understand wrapped tokens

shamisem:


jorge-campo (changes requested):
Hey, @shamisem - Thank you; great job on this article!

I'm reviewing the content with the eyes of someone with only basic crypto knowledge, so bear with me.

## Terminology

We're introducing a few new topics (such as 'blockchain', 'DeFi' apps or 'vault') that we need to elaborate on. I wonder if this is not the time to use the Material for MkDocs glossary option. You proposed that some weeks ago, and I wanted to avoid adding more options to our workflow, but this content makes me think again. I tested Material for MkDocs tooltips a while ago, but I don't remember the maximum glossary visible length (if any) or if we can use more than one paragraph (if we need to).

Could you please check if using a glossary makes sense and looks as expected? [Material distinguishes between tooltips and glossary](https://squidfunk.github.io/mkdocs-material/reference/tooltips/); a glossary is what we want because we can centralize terms on a single file. We should test how this works with a short paragraph. We should also test how it looks on your phone (you can push the article to dev-help.status.im and test them). Please, create a new issue to track this down.

**Can we postpone publishing this article until we find out if a glossary is necessary?**

## Review

- I suggest adding a link to the Wikipedia article explaining a blockchain or adding this to the list of terms in the glossary I proposed to implement.
- In your admonition, "Status Wallet" is a good place to link to the intro Status Wallet article.
- Although you explain "why" I need a wrapped token at the beginning of the article, I'd include this in a bullet point under "The basics": "You can use wrapped tokens to ..." This is, for me, the crucial question the article should answer: why do I need these tokens in the context of the Status Wallet?
- "Wrapped tokens are backed 1:1 by the asset they depend on." This is clear enough, but an example would help. Something like "For example, if you own two WETH, you own the equivalent ETH value.
- XMR (Monero) - Please check with John if we want to use this example. I guess it's OK because your point is about how the wrapped version misses Monero's privacy option. I would add that Monero uses a different blockchain than Ethereum (highlighting again that this is used by the Status Wallet), so it's clear you need its wrapped version in the Status Wallet.
- "your transaction history for XWMR will be public" It's worth clarifying what "public" means with a tooltip. 
- "To do that, go to Wallet > Swap." Please, replace this with the link to the 'swap' article. Imagine that, as a newcomer to crypto; I go to Wallet > Swap. Now, what? :) Let's keep procedural steps in procedures unless we discuss something trivial.

shamisem:
@jorge-campo Thanks for your review!

I added a glossary to the dev branch, so this article will have tooltips once it's merged.

I also added everything else you suggested, including a tooltip explaining what a 'public' transaction history is. We'll sometimes need to use tooltips as opposed to glossary terms, as we don't want this definition to show up every time we say 'public' somewhere in the docs.

 > I would add that Monero uses a different blockchain than Ethereum (highlighting again that this is used by the Status Wallet), so it's clear you need its wrapped version in the Status Wallet.
I reworded my original sentences to 'For example, the XMR (Monero) coin's _original_ blockchain guarantees <...>. Status Wallet is only compatible with the Ethereum blockchain, so you need to use WXMR.' so that it's clear that those are two different blockchains.

jorge-campo:
Thanks, @shamisem ! Great stuff :)

>  including a tooltip explaining what a 'public' transaction history is. We'll sometimes need to use tooltips instead of glossary terms, as we don't want this definition to show up every time we say 'public' somewhere in the docs.

This is interesting for the rest of the team. What do you think about showing two examples of code with tooltips and glossary terms and the reason for using one or the other? You can do this in any of our weekly meetings.

PR #509, opened 2023-03-14
Send crypto to your Status Wallet

shamisem:
Hey @jorge-campo, sending this tiny article for your review!

I don't think we need any cross-linking here, as this article's basically redirecting the reader to another one, meaning extra links may create extra confusion.

### DIFF
--- a/docs/en/status-wallet/send-crypto-to-your-status-wallet.md
+++ b/docs/en/status-wallet/send-crypto-to-your-status-wallet.md
@@ -8,6 +8,11 @@ hide:
 
 # Send crypto to your Status Wallet
 
-To transfer your crypto to your Status Wallet from an exchange or another wallet app, just send it to your Status Wallet address. For more details, check out [Receive crypto][receive-crypto].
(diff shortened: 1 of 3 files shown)


docs/en/status-wallet/send-crypto-to-your-status-wallet.md
jorge-campo, line 1: Hey @shamisem - Two minor things.

1. What do you think about this:

"... just send it to your Status Wallet address _following the instructions or your exchange or wallet app._" 

The problem is, with this proposal, the sentence goes beyond the 25 words limit, which I want to avoid. However, it would be helpful to reinforce the idea that we don't provide instructions about what the user should do with their exchange or wallet app. You could break the sentence into two.

2. The user landing on this article may want to send crypto from their Status Wallet. Therefore, I'd include a final tip admonition pointing to your "Send crypto to someone" article.

shamisem: Hey Jorge, thanks - great points! I've added both to the article.

PR #510, opened 2023-03-14
Restore your Status Community

jorge-campo:
- 459 First draft -- ready for review
- 495 fix typo


docs/en/status-communities/restore-your-status-community.md
```diff
@@ -0,0 +1,45 @@
+---
+id: 495
+revision: 0
+language: en
+hide:
+  - navigation
+---
+
+# Restore your Status Community
```
cheny0, line 9: "Restore your Status Community" sounds like the community is suspended for some reason. How about "Restore access to your Status Community" or "Manage your community on a different device"?

jorge-campo: Thank you, Chen! It wasn't easy to find the name for this article because they way you "restore" a community (as a community owner) is the same way you "import" a community (as a standard user or community owner).

Let me explain why I think "restore" still makes sense:

- This "restore" article follows the "back up" one. Back up and restore are actions commonly used together when discussing technical topics.
- "Restore access to your Status Community" can be misleading and make readers think the access was blocked or suspended (using token-based permission, for example). This is not what's happening. Your community doesn't exist in one device, and you want to restore it on that device.
- "Manage your community on a different device" can also be confusing because you can restore a community on the same device you used the community before. It's not mandatory to use a different device to restore the community.

PR #516, opened 2023-03-17
Remove a Status profile

cheny0:


jorge-campo (changes requested):
Great article, @cheny0 - I like that you include information about what happens on others' devices when I remove my profile.

Here are my review comments:

- If I see the "Delete your Status profile" and "Remove a Status profile" side-by-side, deciding which article is what can be challenging. I think the title in "Remove" needs to be longer to make a difference from "Delete". For example, "Remove a Status profile from your profile list".
- Because of the previous point, the admonition should read, "A profile removed from your device is not deleted ..."
- "... on all devices you use the profile." -> "... on all the devices where you use the profile." 

## "What to expect" section

- We need to avoid sublists within lists. Except in cases like the one we discussed in the "Sync your profile across devices", sublists require readers to pause and analyze the content, which goes against their quick scan process when reading documentation. Please, write each sublist point as a simple list point.
- I think the two cases you mention in the sublist ("messages you sent accessible" and "people see your information") belong to different articles. Let me explain, and feel free to ping me for clarification.
	- The "messages you sent accessible" is something that you need to mention in the "Delete profile" article, but not here. Here, on the "Remove profile" article, I only want to remove the profile from my device, but it's clear this doesn't delete my profile (based on your admonition). Reading this sentence in this article makes me doubt the article's purpose. In other words, the "messages you sent accessible" case is only relevant in the "delete" action.
	- "people see your information" is relevant here, but I think it needs clarification. What do you think about "*Because a profile removed from your device is not deleted*, people still can see all the information you ... " ?
- "You can no longer use a profile if you don't log into it for more than 30 days." I think we should explain this more.
	- Remember that for users, their profile includes their wallet accounts. We must reassure readers that their crypto is secure when discussing removing or deleting information.
	- The sentence, as written, can make the reader think their profile is lost if they don't use it in *one* device for 30 days. But what if s/he uses the same profile on *another* device? It's worth to clarify we mean "not using the profile on *any* device".
	- When you discuss dates (7 days, 30 days, etc.), it's always helpful to say "or more", except if you mean an exact amount of days (no more and no less). Although the reader can implicitly understand this, it is nevertheless helpful.
	- Finally, when you have to give bad news to users, try to provide them with a "way out" :) Your profile is lost but hey, you can always create a new one.
	- I would say, "If you don't log into your Status profile for 30 days or more (on any device), your crypto funds are safe, but your profile is irrecoverably lost. You can always create another Status profile [using your recovery phrase][link]".

## Procedure

- "... then, tap your profile *picture*" -> We use "profile *photo*". See https://write.status.im/terminology/#status-products-and-features-names
- "... if you see a pop-up to sign in." -> "... if you see a pop-up to *menu* to sign in. See https://write.status.im/style-conventions/#general-guidelines.
- "6. Click **Remove Profile**." -> What do you think about "6.  Click **Remove Profile** *to confirm*." to use the same expression you use on mobile?

cheny0:
> Great article, @cheny0 - I like that you include information about what happens on others' devices when I remove my profile.
> 
> Here are my review comments:
> 
>     * If I see the "Delete your Status profile" and "Remove a Status profile" side-by-side, deciding which article is what can be challenging. I think the title in "Remove" needs to be longer to make a difference from "Delete". For example, "Remove a Status profile from your profile list".
> 
>     * Because of the previous point, the admonition should read, "A profile removed from your device is not deleted ..."
> 
>     * "... on all devices you use the profile." -> "... on all the devices where you use the profile."
> 
> 
> ## "What to expect" section
> 
>     * We need to avoid sublists within lists. Except in cases like the one we discussed in the "Sync your profile across devices", sublists require readers to pause and analyze the content, which goes against their quick scan process when reading documentation. Please, write each sublist point as a simple list point.
> 
>     * I think the two cases you mention in the sublist ("messages you sent accessible" and "people see your information") belong to different articles. Let me explain, and feel free to ping me for clarification.
>       
>       * The "messages you sent accessible" is something that you need to mention in the "Delete profile" article, but not here. Here, on the "Remove profile" article, I only want to remove the profile from my device, but it's clear this doesn't delete my profile (based on your admonition). Reading this sentence in this article makes me doubt the article's purpose. In other words, the "messages you sent accessible" case is only relevant in the "delete" action.
>       * "people see your information" is relevant here, but I think it needs clarification. What do you think about "_Because a profile removed from your device is not deleted_, people still can see all the information you ... " ?
> 
>     * "You can no longer use a profile if you don't log into it for more than 30 days." I think we should explain this more.
>       
>       * Remember that for users, their profile includes their wallet accounts. We must reassure readers that their crypto is secure when discussing removing or deleting information.
>       * The sentence, as written, can make the reader think their profile is lost if they don't use it in _one_ device for 30 days. But what if s/he uses the same profile on _another_ device? It's worth to clarify we mean "not using the profile on _any_ device".
>       * When you discuss dates (7 days, 30 days, etc.), it's always helpful to say "or more", except if you mean an exact amount of days (no more and no less). Although the reader can implicitly understand this, it is nevertheless helpful.
>       * Finally, when you have to give bad news to users, try to provide them with a "way out" :) Your profile is lost but hey, you can always create a new one.
>       * I would say, "If you don't log into your Status profile for 30 days or more (on any device), your crypto funds are safe, but your profile is irrecoverably lost. You can always create another Status profile [using your recovery phrase][link]".
> 
> 
> ## Procedure
> 
>     * "... then, tap your profile _picture_" -> We use "profile _photo_". See https://write.status.im/terminology/#status-products-and-features-names
> 
>     * "... if you see a pop-up to sign in." -> "... if you see a pop-up to _menu_ to sign in. See https://write.status.im/style-conventions/#general-guidelines.
> 
>     * "6. Click **Remove Profile**." -> What do you think about "6.  Click **Remove Profile** _to confirm_." to use the same expression you use on mobile?

Thanks @jorge-campo! Re-edited and merged. 

Adding conclusions of our further discussion on discord here:
1. Move "messages you sent accessible" and "people see your information" to "Delete your Status profile"
2. Add [delete-a-profile] after [irrecoverably lost] in the ##What to expect list.

PR #517, opened 2023-03-17
Delete your Status profile

cheny0:


jorge-campo (changes requested):
Thank you, Chen! I like the list of things the user should take into account. I didn't think about that, but it makes total sense.

## Content

### Intro
- We should clarify more about how the Waku p2p network works. I would start the article by saying, "Status uses a peer-to-peer network that temporarily stores your profile information and data. Because of this decentralized network, there is no one-click ..."
- This article works best with a short "What to expect" H2 list collecting the main points. These main points are:
	- The only way to delete your Status profile is to stop using it for 30 days or more.
	- Once your profile is deleted, it's irretrievably lost. You can always create a new Status profile.
	- Deleting your Status profile doesn't remove your wallet funds. Your wallet funds are safe and available if you know [your recovery phrase] [create-or-restore-your-status-profile-using-a-recovery-phrase].
- It's important to clarify that your wallet funds are always safe if you have your recovery key. This might be users' main fear when thinking about deleting their profile. Therefore, I would add this information in the "What to expect" list.

### List
- Since we use a previous "What to expect" H2, we need a new "Delete your Status Profile" H2 right before "Here are things you can ..."
- "Transfer assets out of your wallet" -> "Transfer assets out of your Status Wallet."
	- I would make this the first element on your list, as it's probably the most important.
- It would be great to style this bullet points list with some nice checkboxes. We need to create a new style for that. In the meantime, please create a new issue for yourself to update this article and link the issue to #520 ?

## Markdown
- You're missing the link in the [peer-to-peer network] reference. The link is [peer-to-peer-messaging], as described in the `urls-en.txt` file.
- You mention an article that doesn't exist yet, "Leave a community". In this case, you need to create the placeholder. This means:
	- Create the entry in the Miro board. In this case, the article goes under Status Communities > Status Communities
	- Create the GitHub issue. Add the GitHub issue number and link to the Miro board entry.
	- Add the placeholder Markdown file to the repo in the corresponding folder. 
	- Add the new article's URL to the `urls-en.txt` file (you don't need to add the hamburger menu entry.)
- Same goes for the "Edit your profile" article.

We can clarify any questions you may have via chat or during our next call.

cheny0:
Thanks @jorge-campo, I applied your comments. Content has been merged.

cheny0:
Adding decisions of our further discussion on Discord here:
1. Add "messages you sent are still accessible" and "people still see your information" in ##What to expect.
2. Avoid the word "delete" and rewrite relevant sentences.

PR #521, opened 2023-03-20
Status Wallet FAQ

earakel:
Hey Jorge, sharing Status Wallet FAQ v1.0 for your review!

### DIFF
--- a/includes/urls-en.txt
+++ b/includes/urls-en.txt
@@ -13,6 +13,7 @@
 [download-status-for-linux]: /en/getting-started/download-status-for-linux
 [download-status-for-mac]: /en/getting-started/download-status-for-mac
 [invite-friends-to-status]: /en/getting-started/invite-friends-to-status
+[new-profile]: /en/getting-started/run-the-status-app-for-the-first-time#new-profile
 [run-the-status-app-for-the-first-time]: /en/getting-started/run-the-status-app-for-the-first-time
 [what-is-status]: /en/getting-started/what-is-status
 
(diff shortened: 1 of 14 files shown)


includes/urls-en.txt
shamisem, line 1: Not sure about why we have three URLs all pointing to the same article, so I removed two of those

jorge-campo (commented):
Thanks, @shamisem ! Great questions and answers. Succinct and super clear.

I have only minor suggestions.

- Is it FAQ or FAQs? I understand the "Q" stands for "questions" (plural), but a few searches on Google indicate the "Qs" form is more common.
- Please, don't use the bold typeface for the whole title. Something like "Status Wallet **FAQ**" works.
- You can include Status Web as well, after mobile and desktop :) Repo is here: [status-im/status-web: Embeddable, customizable and themable component for your Community. (github.com)](https://github.com/status-im/status-web).
- "Status Wallet can only be used for Ethereum-based assets." -> What do you think about using the active form? "You can only use Status Wallet for Ethereum-based assets".
- "This includes Ether (ETH) and ERC-20 tokens, ..." -> Maybe "This includes Ether (ETH) and *other* ERC-20 tokens, ..." so it's clear ETH is an ERC-20 token.
- One of the first questions a user may have is, "is Status Wallet secure?". It's difficult to answer because, although we open-source our code and make everything possible to keep users' funds secure, nothing is 100% secure. User actions can also make the most secure software unsafe. We can use the following information to clarify this question:
	- Status Wallet is open source and non-custodial.
	- We don't use third-party services to store public or private keys.
	- Your keys are encrypted by your Status password (that only you should know).
	- Transactions only occur after they're confirmed on the device.

PR #533, opened 2023-03-24
Understand network fees 

shamisem:


jorge-campo (changes requested):
Thanks for your work, Emma - I'm afraid we need to rethink and rework this article. It relies too much on Ethereum PoW concepts that are no longer valid under the PoS consensus mechanism.

I propose you the following changes:

- Since "gas fees" is a widely accepted term, I will introduce this earlier. "Network fees are charged by ..." -> "Network fees (also known as gas fees) are charged by ..."
- About the admonition. "Status Wallet doesn't charge any extra transaction fees." We need to clarify this. While it's true that the Wallet doesn't charge fees, Status will charge fees when the user buys crypto using it (buying crypto is also a transaction, from the user's point of view). It's worth clarifying if there is any other case where we charge fees (not that I'm aware of). This is a question for John, and we should include this information in the article to be completely transparent.
- "Status Wallet doesn't charge any extra transaction fees." See my previous comment. What about "Status Wallet doesn't charge any extra *network* fees."?
- "You can set custom network fees for your ..." You can link here to your other article, "Handle pending or stuck transactions".
- "When you send a blockchain transaction, it needs to be processed and validated by the network." I would slightly change this to "When you send a ~~blockchain~~ transaction, it needs to be processed and validated by the *blockchain* network." Your original sentence can make the user think there are transactions other than blockchain transactions to do with their Status Wallet.
- "This involves the work of validators, who use powerful computers to validate transactions." -> "This involves the work of validators, who use ~~powerful~~ *their* computers to validate transactions." In Ethereum PoS, you no longer need a powerful computer to run transactions. Some implementations, like Nimbus, can be configured in a [Raspberry Pi](https://nimbus.guide/pi-guide.html).
- "By sending a transaction, you create a job validators compete for." This is no longer true in Ethereum PoS. Validators don't "compete" with each other. Validators are selected through a random process, and once they are selected, they create the block and include transactions in it. I won't go further into the details here because we'll have a dedicated article to explain the blockchain.
- "A network fee's size depends on how busy the blockchain is, as well as the size of your transaction." -> "A network fee's size depends on how busy the blockchain is, as well as the ~~size~~ *complexity* of your transaction." Transacting higher amounts of crypto doesn't necessarily mean you'll pay more fees. It depends on the complexity of your transaction; more complexity means more resources and fees.
- "Network fees are measured in gas, which in turn is measured in gwei." While this is correct, it makes things harder to understand. Network fees and gas fees are equivalent terms. I would say, "Network fees are measured in gwei."
- "If you set your network fees too low, it may take longer for your transaction to be verified by the network. In some cases, it may never be processed at all." Again, this is the perfect place to link your other article, "Handle pending or stuck transactions".
- "Blockchain transactions can be compared to job offers ..." Your analogy applies more to a PoW scenario. You can use ChatGPT to find an easy-to-understand analogy for the PoS scenario.
- "Validators invest their time and resources into processing transactions, which includes using computer hardware that consumes electricity". Again, this doesn't apply to the Ethereum PoS consensus mechanism. The main goal of network fees is not to cover the validators' electricity bill but to incentivize validators for their work and contribution to the blockchain.

PR #550, opened 2023-04-06
Delete your wallet

shamisem:


jorge-campo (changes requested):
Thanks, @shamisem !

The content makes easy to undersand the concept. However, it needs an important part of the wallet: accounts.

In Status Wallet, users have one by-default wallet account, but they can create additional ones. A typical use case is not to delete my entire wallet but to delete one or more accounts I don't longer want to use.

We should mention this and adapt the wording to apply to the whole wallet and one or more wallet accounts.

Considering this, change the title from "Delete your wallet" to "Delete your Status Wallet or wallet account". Even when your text applies to any crypto wallet, we want readers to associate our docs with our app.
- From a visual point of view, the content can be easier to scan if you "break" the text blocks with an admonition. It's a simple trick, but the article looks "easier". In this case, I'd use the "Remember that ..." inside and tip-style admonition, and place the admonition right before "If you no longer ..."
- After the last sentence, "... and uninstall the app.", mention that the user can always recover the wallet or wallet account as long as they know the recovery phrase. Then, you can link to the article explaining how to import an account. 
- "... send your leftover crypto to some other wallet address." What do you think about this? "... send your leftover crypto to some other account or wallet address."
- You're missing the `status-wallet/index.md` entry for this topic.

PR #551, opened 2023-04-10
About the Community History Service

cheny0:


jorge-campo (changes requested):
Thanks, @cheny0 ! Here are my review comments.

I think we need to revise the article. After reading it, as a new user, I still have questions:

- It's unclear how the CHS works in the background.
- It's also unclear if this service runs only on the owner's computer or on any community member. (I think we can detail more the magnet link example with Alice.) You mention that the service also requires members, which makes me think the community members also run a BitTorrent client on their computers. If this is true, we need to be more specific about that.
- Without getting into technicalities, we need to explain better how this works for the community member and the owner.

The information is there, but it reads like it's spread across the article, and I need to read everything to understand how the whole thing works. Organize the information in blocks so the reader moves from one block to the next with all the information from the previous block. By "blocks" I mean:

- What's the Community History Service?
- What are the benefits?
- What does it mean for the community owner; what happens on their computer?
- What does it mean for members; what happens on their devices?
- How to keep the service running?

## Intro

- Status uses a [peer-to-peer network](http://localhost:8000/en/using-status/about-status-messages#peer-to-peer-messaging) to temporarily store your community messages for 30 days.
-> What do you think of "Status uses a [peer-to-peer network](http://localhost:8000/en/using-status/about-status-messages#peer-to-peer-messaging) ~~to~~ *which* temporarily store your community messages for 30 days."? The "to" preposition makes you think that the goal of the p2p network is to store messages, but the network also has other functions.
- "... messages are automatically downloaded and stored on their computers locally." -> "messages are automatically downloaded and *locally* stored on their devices ~~computers locally~~." The device doesn't need to be a computer; the same process occurs for community members using the community on their phones.
- "After 30 days, the messages don't exist on the peer-to-peer network, ..." I think this needs clarification. This may make one think all messages are removed after 30 days, no matter how old they are. We want to say that only messages older than 30 days are removed. -> "Messages older than 30 days are no longer stored in the peer-to-peer network, ...", or the active form "The peer-to-peer network purges messages older than 30 days, ..."
- I would split the introductory info into two different paragraphs. The first paragraph explains one idea, the problem, and ends with "... can't see these messages." The second paragraph explains another idea, the solution, and starts with "With the Community History Service (CHS)..."

## The basics

- You talked about "storing historical messages" in the introduction, and you now start to talk about "backup". I would avoid the term "backup" because it makes one think your community messages are stored elsewhere and you can restore them whenever needed. This is not what the CHS provides. I would avoid the word backup. For example, in the firsts bullet point, I'd say, "Status is not involved in storing your community messages and doesn't have access to them. Messages older than 30 days are stored in your community node." (or something similar; feel free to adjust this text).
- "The backups are stored on your computer and the computers of your members." Apart
- "During the first download, members ..." What's the "first download"? Why do members need to download messages? Isn't this a service relying solely on the community owner's computer? This sentence needs clarification.
- We need to mention that the information is encrypted, so I can't access its contents even when it's on my computer through the BitTorrent client.

## How it works

- The H2 title should use "Community History Service" in uppercase letters, as this is a unique Status app feature.
- We must emphasize that, as long as the CHS is not turned off, all these processes in the community node and members' computers occur automatically. In other words, there is no need for the owner or members to do anything for this to work.
- It isn't clear where the CHS runs. You mention the owner and the member, but it's unclear what role each one plays in the process. A checklist of actions right before Alice's example would help to understand what's going on. Something like:
	-  [x] The CHS downloads messages older than 30 days on the owner's computer.
	-  [x] CHS creates magnet links for ...
	-  [x] A member wants to check for a channel conversation over 30 days.
	-  [x]  Etc.
- You can use a tooltip to describe what a magnet link is. I bet users are not familiarized with them, even if they've used BitTorrent in the past.

## Maintain the CHS's stability

- I think "availability" works better than "stability" in this title. By not following basic rules with CHS, the service becomes unavailable. "Available" and "unavailable" are common words when describing computer services.
- "It is ideal ..." We need to be more assertive here. Something like "To provide the best experience for your members, keep Status desktop online every day, or at least once every six days."
- "Don't delete backup files ..." Again, let's not use the word "backup". Let's simply say that users should not remove or manipulate .torrent files as they're part of the CHS mechanism.

cheny0:
Superseded by #588

PR #552, opened 2023-04-10
Turn off the Community History Service

cheny0:


jorge-campo (changes requested):
Thanks, @cheny0 !

My comments for this article are similar to the ones I mentioned on the CHS concept topic.

- Admonition: "... the community history service..." -> "... the Community History Service..."
- "... or when you run out of storage space" -> what do you think of "... or when your computer runs out of storage space" ? This way, we clearly refer to the storage in your computer and not somewhere else.
- You mention "to protect community member's IP addresses." Still, you need to provide more information about what happens with these IP addresses. Add this information to the concept topic for CHS, and backlink to the concept topic in the first paragraph. A topic intro is the perfect place to link to other topics. Always think about what link(s) you can include, allowing readers to jump to other topics they feel more interesting in.
- "New members can't ..." Are you sure this applies to new members only? 
- "... can no longer download the existing backups". I have here the same issue I had while reviewing the CHS concept topic. It isn't clear what backup clients download while the service is on. This is not the place to clarify this, it's the concept topic, but you can add some extra information here.
- I believe Pascal mentioned that once you turn it off, you can't turn it on again (correct me if I'm wrong). In this case, we should inform the owner about this. It's an important limitation and a good reason to avoid turning it off.
- Turning off the CHS degrades the member's experience considerably. Add an admonition suggesting to keep this option on (you can use the same admonition on the other CHS concept topic).

cheny0:
> 

Thanks, @jorge-campo. I have applied the comments and published the article.

PR #555, opened 2023-04-11
490 Understand token requirements in communities

Fabiomorais87:
Hi @jorge-campo,

Fabio


jorge-campo (changes requested):
Hey @Fabiomorais87 , here is my review:

- In your second paragraph, remove "Community owners can set up token requirements for their communities. To [join a community](http://localhost:8000/en/status-communities/join-a-status-community), you need to meet its token requirements." You've already discussed this information in the previous paragraph; no need to repeat us here. You can move the "join a community" link to the first paragraph as well.
- Don't pluralize token names: It's "10 SNT" and not "10 SNTs".
- "... as a requirement to join a community" -> "... as a requirement to join *their* community."
- "In that case, you need to have a minimum of 10 SNTs to join that community." -> "In that case, you need ~to have~ a minimum of 10 SNT *in your [Status Wallet][link-to-the-status-wallet-article]* to join that community." It may not be immediately obvious to the user where those tokens should be to validate your eligibility.
- "There are three kinds of tokens in the Status app: assets, collectibles, and ENS" -> You have an Oxford comma before "and" that you need to remove (as funny as it sounds, Oxford spelling doesn't use Oxford commas).
- I think we can improve a bit the note admonition:
	- "When you join a private community, the token required to join is not transferred to anyone. They remain on your account. They are simply used for verification purposes, and your tokens remain in your possession."
	- -> "When you join a private community, the token or tokens required to join are not transferred to anyone. They are used for verification purposes only, and they remain in your possession and wallet."
- "You meet the requirements" -> I believe these titles work better with the "if", "*If* you meet the *token* requirements."
- "You don't meet the token requirements" -> "*If* you don't meet the token requirements."
- The "meet requirements" / "don't meet requirements" need screenshots. Please, ask Francesca to create them for you and if it's OK to use them already. This article's readability would increase significantly.
- In the "You don't meet the token requirements," you don't explain to users how they should fix this condition. You say that they need to have the required tokens, but how? This is the perfect moment for a call to action. To buy the required tokens, read [Buy crypto][link-to-emma's-buy-crypto-subsection].
- The placeholder article "Tokens, collectibles, and ENS names" has an Oxford comma you don't need.

After these changes, the article is ready to go 🚀

Thanks!

PR #557, opened 2023-04-11
Understand your wallet balance

earakel:
Update: this article still needs a screenshot from Francesca, but otherwise it's ready for review.

### DIFF
--- a/config/en/mkdocs.yml
+++ b/config/en/mkdocs.yml
@@ -213,6 +213,7 @@ nav:
     - status-wallet/index.md
     - Getting started:
       - Status Wallet FAQ: status-wallet/status-wallet-faq.md
+      - Understand your wallet balance: status-wallet/understand-your-wallet-balance.md
     - Send and receive crypto:
       - Send crypto to your Status Wallet: status-wallet/send-crypto-to-your-status-wallet.md
     - Understand crypto assets:
--- a/docs/en/status-wallet/handle-transactions-you-didn-t-make.md
+++ b/docs/en/status-wallet/handle-transactions-you-didn-t-make.md
@@ -1,0 +1,16 @@
+---
+id: 556
+revision: 0
+language: en
+hide:
+  - navigation
+---
+
+# Handle transactions you didn't make
+
+ :octicons-tools-24: In progress
+
+!!! note ""
+     We're working on this content.
+
+--8<-- "includes/urls-en.txt"
--- a/docs/en/status-wallet/index.md
+++ b/docs/en/status-wallet/index.md
@@ -10,6 +10,7 @@ Status Wallet is an open-source non-custodial wallet that helps you store, send,
 ## Get started
 
 - [Status Wallet **FAQ**][status-wallet-faq]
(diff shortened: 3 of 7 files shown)


jorge-campo (changes requested):
Thanks for your article, @shamisem !

We need to add some changes. Here is my review.

- In your title, "Wallet" should use uppercase "W". Even when your article works for any wallet, we discuss the Status Wallet here, not just any wallet.
- You're missing the link for this article at `urls-en.txt`. Please, check the article itself, the `index.md` entry, and the hamburger menu entry with the Material for MkDocs preview. Alternatively, push your article to `develop` to check for these.
- ~~There are more analogies to bank apps in crypto wallets than differences. Similarly to bank apps, your money is not on the app; it's on the bank. Everyone understands that if they uninstall the bank app, the money isn't gone, but many don't understand that if they uninstall the wallet app, crypto is not gone.~~

~~It would be easier for a crypto newcomer to have a bank app reference, even when crypto has little to do with banks. Instead of using the "bank app" example to tell users "we're not the same", why not use this example to welcome them with terms they are already familiar with?  It's a better (and more welcoming) use of your example.~~

> **Note**: (update Jun, 22.) After our conversation, we've decided that using the bank analogy for discussing how a crypto wallet works is not a good idea.

Besides, I find the "Crypto wallets aren't bank apps" sentence a bit discouraging/negative.

- I would clarify what's "fiat". I've seen many newcomers asking this question in forums. Maybe you can do that right after talking about EUR/USD.

- "To double-check...". I wouldn't write this. If we tell users they can "double-check" the balance we present them in their wallet, we're implicitly saying they can't fully trust Status Wallet. And this is not the case. What I would discuss in this section is the fact that the balance fluctuates because the relative crypto-fiat value fluctuates in the market.

- "you didn't make", I would use "you don't recognize". Sometimes you make transactions you don't recognize or remember, but they are transactions you actually made.

- You mentioned the missing screenshot, but you can add the information for that screenshot now. Think of your article with a screenshot like the one I attach, and explain what total balance (A), account balance (B), asset balance (C), and balance changes (D) are. For this, use a tabbed interface, like the one you can see at https://help.status.im/en/status-communities/channels-your-quick-start-guide/#__tabbed_1_1 (2)

balance

---

Screenshot 2023-05-05 at 9 37 14


This way, your article will be ready for the screenshot, and you simply need to include it once it's ready. Concept-type articles should be related as much as possible to the Status app. Try to use examples/screens from the app that helps the user to relate this content to our app.

I'm happy to review this offline with you if that helps.

Thank you.

shamisem:
Thanks for your review Jorge! This branch is quite old, so I feel it would be easier to simply close it and create a new one than fix all of those merge conflicts. I will apply your comments and follow up with the new version.

jorge-campo:
Superseded by https://github.com/status-im/help.status.im/issues/463.

(hey @shamisem, when you replace a branch or PR with another one, add a closing comment like the one above. This helps to track things down when necessary. Thank you.)

PR #583, opened 2023-04-25
Receive crypto

shamisem:
Hey @jorge-campo! 
This article will need a screenshot from Francesca, but otherwise it's ready for review.


jorge-campo (changes requested):
Hey @shamisem - Here is my review.

Your article's structure is atypical and doesn't follow the [style guide structure for a procedure](https://write.status.im/structuring-the-content/#procedure-structure). However, I can understand the reason for that. On the one hand, placing the task at the end of the article won't work because there is plenty of content before. On the other hand, I see this content as a requirement. Therefore, let's consider this an exception without forgetting that, whenever possible, topics should adhere to the structure described in the style guide.

- "(which starts with `0x`)". Good point. I find it useful to provide information on how to identify your public key. However, "0x" just means hexadecimal, and some wallets might include the "0x" prefix to indicate that the private key is in hexadecimal format, even though this is not a standard practice.
What do you think if, in addition to the prefix rule, we also discuss the key length rule? -> 64 characters for private keys and 42 characters, including the "0x" prefix, for public addresses.

- Mobile > Step 2: "From the _navigation bar_, tap..." -> "From the _top navigation_, tap..." (see [style guide UI areas for mobile](https://write.status.im/terminology/#status-mobile-user-interface-areas)).
- What do you think about the following two steps?
    - Step 4. To copy your public address, click :desktop-copy: Copy address. Send the address to the sender.
    - Step 5. Alternatively, ask the sender to scan your public address QR code with their crypto wallet.
    - I include step 5 because we need to consider the case that sender and receiver are next to each other and, in this case, scanning the QR is the fastest (and generally more secure) way to share the address.

- I repeat here the suggestions I was discussing in our group chat:
  - Under "Your legacy and multichain addresses", when you mention "click Edit", I'd mention where this screen is. You say that in the preceding procedure, but it's nevertheless useful. This will force you to rephrase things so sentences are not too long.
  - Many people (even people familiar with crypto) have never used these multichain addresses. Users may think, "If I share this multichain address, will the sender be able to use it?" Maybe it's good to explain that the sender wallet must also support multichain (?)

Thanks.

shamisem:
Hey @jorge-campo, thanks so much for your review now and for your suggestions earlier!

1. Regarding the structure: all true, I struggled quite a bit with organising this. The regular structure didn't work for the reason you've just pointed out - there's _way_ too much that needs to be explained to cover possible questions, but it's possible that plenty of users don't care about the details and just want their mainnet address. Pushing the procedure outside of the first screen could give the impression that it's not even there, which why I ended up moving that section to the very top.
1. `0x`: solid suggestion, all done.
1. Mobile > Step 2: done.
1. Mentioning the QR code: makes perfect sense, done. It's not an extra step, though, so I added this info to the fourth one to avoid confusing the user.
1. 'Your legacy and multichain addresses', explaining where the 'Edit' screen is: took me a tad longer to figure out the best way to do this, so here goes. All done!
1. On the sender's address supporting multichain: I added this note (line 38) earlier following your suggestion on Discord:
`To use your multi-chain address, the sender needs to use a wallet app that supports multi-chain addresses. Status Wallet does this by default.`

Sorry for not responding to your suggestions on Discord earlier - I wanted to do so after implementing them and then got stuck with the 'Your legacy and multichain addresses' section.

jorge-campo:
No need to be sorry, @shamisem ! All good 👌 

Awesome work on a topic that it's not particularly easy to explain to newcomers. I'm not a newcomer, and I also learned something new! :smile:

PR #601, opened 2023-05-08
585 Customize community and channel notifications

Fabiomorais87:
Hi @jorge-campo,

Fabio

### DIFF
--- a/docs/en/your-profile-and-preferences/customize-community-and-channel-notifications.md
+++ b/docs/en/your-profile-and-preferences/customize-community-and-channel-notifications.md
@@ -36,7 +36,7 @@ You can turn off your notifications for specific communities for a designated pe
     1. Choose the community you want to mute and click :desktop-notification: **Mute**.
 
 !!! tip
-	To turn notifications back on, open the muted community, then tap :mobile-more-options: **More options** > :mobile-unmute: **Unmute community".
+	To turn notifications back on, open the muted community, then tap :mobile-more-options: **More options** > :mobile-unmute: **Unmute community**.
 
 ## Mute community notifications except global mentions
 


jorge-campo (changes requested):
Thank you, @Fabiomorais87 - Here are my review comments.

**Intro** 

Notifications you set for communities apply to channels as well. However, this may not be clear for a first-time user. The article's title includes 'channels', but then there is no mention of channels in the topic. This may be confusing for a first-time user. I suggest you include a brief explanation. You can also use this explanation to introduce the 'exceptions' topic. Something like this may work: 

> The notifications you customize for a community apply to all its channels. For example, suppose you decide to [mute all community notifications][#mute-all-community-notifications]. In that case, you won't receive any notification from any channel, even if someone mentions you. To customize notifications for specific channels, check out Set up and customize notification exceptions.  

**Mute all community notifications.** 

- 'It starts appearing greyed out'. We don't say how things start or end appearing; we describe how they are. You can simply say, 'When you mute all notifications for a community, a :desktop-mute: icon appears next to the community name.' 

- Mobile section. In the previous sentence, the :desktop-mute: icon is the one with the bell crossed; it's not the one you use in the admonition. If the community is already muted, the bell icon is crossed. 

- Mobile section: It's not 'Mute', it's 'Mute community' (see image) 

Screenshot 2023-06-05 at 3 18 36@2x

- You can remove the tip; you wrote the same information in the introduction of this H2 section. The tip you need there is the one that explains to the user how to turn the community notifications on again when they choose 'Until you turn it back on'. 

**Mute community notifications except global mentions** 

Let's explain, if only briefly, what's the outcome of this action. After your first sentence introducing the procedure, you can say, 'This example explains how to turn off all notifications for a specific community while maintaining global mentions. This can be helpful in very active communities where you're only interested in global mentions, such as @everyone'. 

- 'If you want to avoid missing important messages, you can selectively mute some notifications, except global mentions.' It is useful here to (1) link to the corresponding 'mentions' topic, where we explain what 'global mentions' are, and (2) define briefly what global mentions are. For example: 'If you want to avoid missing general announcements, you can selectively mute some notifications, except global mentions (such as @everyone). 

- Mobile section. The dark bell icon you use is not the one for the Notifications settings option. See the image below. 

- 'MAke sure' -> 'Make sure' 

- Add a tip admonition explaining to the user that s/he can follow the same procedure to keep only personal notifications active. The procedure is about global mention, but we can always include this information as a tip. 

**Seet the same settings for all community message types** 

- H2 title: I think the title should mention 'notifications' -> _Set the same notification setting for all messages in a community._ 

- 'Currently, you can only set the same settings for all community types using Status mobile.' -> 'Currently, you can only set the same settings for all community _message_ types using Status mobile.' 

- The 'Notification settings' icon is not correct.

Your final admonition can improve a bit: _You can customize notifications for channels in a community using exceptions. For more details, check out Set up and customize notification exceptions._

**Accessibility** 

There are two options to complete the procedures you describe: long-press the community ('long press' option) or tap the community and tap the more options button ('tap twice'). After a second thought and gathering more information, I think it's better we describe the tap twice option and not the long-press one. The reason is that elderlies, people less familiar with technology, and people with motor control challenges can find the 'long press' option harder to complete, and they may not be aware there is another option as we only document one of them. 

Please, let me know what you think and, if you agree, update the procedures to describe the 'tap twice' option instead. You don't need to increase the number of steps, as you can describe the two actions in the same step. For example: x. Tap the community you want to mute and, in the top navigation, tap :mobile-more-options: More options.

PR #609, opened 2023-05-11
597 Customize direct messages notifications

Fabiomorais87:
Hi @jorge-campo,

Fabio

### DIFF
--- a/docs/en/your-profile-and-preferences/customize-direct-message-notifications.md
+++ b/docs/en/your-profile-and-preferences/customize-direct-message-notifications.md
@@ -39,7 +39,8 @@ This action mutes notifications from all your direct messages (DMs). If you want
     1. From the tab bar, tap :mobile-messages: **Messages**.
     1. Tap your profile picture.
     1. Scroll down the list and tap :mobile-mute: **Notifications**.
-    1. Turn **Notifications** off.
+    1. Select **Direct messages** > **Your default notification settings**
+    1. Turn off **All messages**.
 
 === "Desktop"
 
@@ -59,7 +60,7 @@ This action mutes notifications from all your direct messages (DMs). If you want
 
     | Preference | Description |
     |:---|:---|
-    | Notifications on | Turn this on if you want to receive direct message notifications. |
+    | Notifications on | Turn this on if you want to receive notifications banners and lock screen notifications for direct message notifications. |
     | In notification centre only | Turn this on to receive direct message notifications in the Status notification centre. |
     | Notifications off | Turn this off if you don't want to receive any direct message notifications. |
 


jorge-campo (changes requested):
Thanks for your content, @Fabiomorais87 – Here are my review comments.

- Title: "Customize direct messages notifications" -> "Customize direct _message_ notifications" (singular "message'). The same goes for all titles and places where you mention "messages notifications"; change them to "_message_ notifications". Both are gramatically correct, but the singular form is more common. I use ChatGPT to find out what's more common and easier for non-native English speakers to understand.

- Let's reword the first paragraph a bit; sentences feel disconnected and unnecessarily long.

    > By default, you receive notifications for direct messages (DMs). You can customize the DM notifications and receive only the ones relevant to you.

- These two paragraphs (1, 2) express the same information and they're next to each other. This looks repetitive, so remove one of them. Because the information is relevant to all the H2 sections you discuss, I'd keep the one in the intro (1) and remove the one in the procedure (2).

Screenshot 2023-06-21 at 10 04 37@2x

- "Long press the DM you want to mute." See my 'Accessibility' comment from previous reviews here https://github.com/status-im/help.status.im/pull/601#pullrequestreview-1462680908. The same goes for the rest of the procedures in the topic where you mention "long press".

- "Select the length of time..." -> "Select the amount of time ..." Lenght is dramatically correct but harder to understand. When in doubt, check with ChatGPT for what's most common for non-native English speakers:

>  "Select the amount of time you'd like to mute your notifications," may be slightly clearer for some readers. It explicitly mentions "amount of time," which provides
 a more specific and straightforward understanding of the action to be taken.

- In the tooltip for the first procedure, use bold typeface for "Mute".

- "If you want to mute all Status notifications..." -> "If you want to mute all _the Status app_ notifications..."

- "Mute all direct message notifications procedure". The procedure you describe mutes ALL notifications, not only the direct message ones. Please, review the procedure at https://www.figma.com/file/JlpPhJp0SMnEyBQy4nOZHo/Settings-for-Mobile?type=design&node-id=3875-347501&t=eqUFY2yjrxI30A87-0

Screenshot 2023-06-21 at 10 38 42@2x

- Your admonition "If you mute all Status app notifications, you're also muting notifications that appear on your screen and disappear shortly after." This admonition doesn't work well here because this procedure is about muting all the _direct message notifications_, not _all the Status app_ notifications.

- The terminology is confusing here. When you mean "notification banners", do you mean the same as "notifications that appear on your screen and disappear shortly after." previously? If yes, use the same name for both expressions. If not, clarify what "notification banners" are.

- "Turn this on to receive subtle notifications without banners. You need to unlock your phone to see your notifications." I don't think this is correct. You need to unlock the phone to _read_ the notifications but not to _see_ the notifications, right? You see the notification on the locked screen, and you still need to unlock the screen to read it. Correct me if I'm wrong.

Thanks.

PR #611, opened 2023-05-12
Respond to a contact request

cheny0:
- Create espond-to-a-contact-request.md
- minor update
- Ready for review
- Add article link to add-a-contact-in-status.md


cheny0:
Need to add contact request option.

jorge-campo (changes requested):
Thank you, @cheny0 – Great content! I have several points but none of them fundamentally change your content.

## Intro

- "Status users with access to your profile can send you contact requests" 
    - Uhm. This is ambiguous and creates uncertainty for the reader. "Who has access to my profile?" or "How can I configure access to my profile?" they may ask. 
    - In reality, anyone except users you have blocked can see your profile and send you a contact request. This will be people you interact with in group chats or communities, but it can also be people who know your public key or profile name and search for you. (Note: in the near future, the "search" field users can see in the app will search for anything: communities, contacts, chats, etc.) 
    - Therefore, I would make this first statement more clear by saying that people you interact with or search for your profile can send you a contact request. Then, I would add a note-type admonition saying that people you have blocked can't see your profile or send you a contact request; and you link this admonition to your "block a user" article. I hope this makes sense; if it doesn't, ping me. 
    - Alternatively, you can integrate the "blocked" information in the paragraph and maintain your tip. 

## "What to expect" section 

This needs some clarification.

- "Contact requests expire after seven days" -> What about "A contact requests someone sends to you expires after seven days." 
- Your bullet point number 3 goes after number 1, from a logical point of view. 
- "You can turn off contact requests to stop receiving new requests" -> "You can turn off contact requests to stop receiving any new request." It's always good to clarify that turning off contact requests will affect all contact requests (any). 

## Respond to a contact request 

Be careful here with the table. It seems to me you're explaining the different UI options, and this is something we don't do. Any time a user taps or clicks on something, they'll be presented with a set of options, but it's not our goal to provide a description to all of them but to guide them towards the solution of a task. We've used tables in other procedures, and that's OK, but in this particular one, the task at hand is clear: I want to accept or reject something.

- In your case, I think you can resolve the last step by saying, "Tap **Accept** or **Decline** the contact request or tap **Review contact request**. If you decline, the requester doesn't receive any notification." 
- You explained the notification thing in the "What to expect" section, but it's an important element and worth repeating. 
- The "tap the sender's profile picture" would be, for me, a tip-type admonition. You can put two admonitions together at the end of a procedure; it's uncommon, but it's OK. Up to you. 
- Great work on that tip; it's exactly where it should be and provides the exact information someone needs at that moment. 

## Stop receiving contact requests 

- "… tap the toggle on the right". Except for pop-up menus and checkboxes, we don't mention the UI element, whether button, toggle or something else. In this case, we use turn [on|off] (see https://write.status.im/style-conventions/#general-guidelines).
- Your sentence would be "In the **Privacy list**, find and turn off the **Allow new contact requests** option."

cheny0:
Thank you @jorge-campo. I have applied your comments and merged the content.

PR #613, opened 2023-05-13
If someone else has access to your Wallet

shamisem:
Hey @jorge-campo, sending this for your review!

This article breaks the rule of no links in procedures when talking about switching between accounts. Otherwise, I'd have to either to list the same steps as Chen does in her article several times, or risk confusing the user by leaving this step out of the procedure - you'll skip admonitions when you're in rush.


jorge-campo (changes requested):
Thank you, Emma - Here are my review comments, mostly about the content structure. I hope this helps.

- This article was written before the new content organization you made. The new sub-section for the article is 'Secure your wallet'. You must `rebase` the branch and fix the merge conflicts before pushing to `master`. Remember also to update the links in the hamburger menu (if they're not updated yet). 

- 'Your Status Wallet and profile share the same recovery phrase'. You can link here to the recovery phrase article, 'Understand your Status keys and recovery phrase'. 

- 'While your profile is protected by a password locally, anyone who has the recovery phrase can log in to your profile even without your password.' What do you think we include the wallet information as well? For example, 'While your profile ~~is~~ _and wallet are protected_ by ~~a~~ _your Status_ password, anyone ~~who has the~~ _with your_ recovery phrase can log in to your profile and wallet even without your password. 

- Right after the first paragraph, you can add a tip-style admonition to the 'Back up and secure your recovery phrase' topic. 

- Your content is complete and follows a logical order; great work on that. However, its structure can improve a bit. Having steps inside steps doesn't make things easy for a reader that, most probably, is in a bad situation. I think the following changes will make things much easier to scan and follow. Try them 😄  You can share the 'before' and 'after' versions and ask in our group chat what the rest of the team thinks. 

    - In our structure, steps work better when they're in H3 but not when they're in H2. Create a single H2 with a name similar to 'What to do if someone else has access to your Wallet'. 

    - Transform Step 1, Step 2, and Step 3 into H3s under the single H2. 

    - Merge your Step 1 and Step 2 H3s' procedural steps under a single 'Step 2' H3. In other words, merge your procedural steps so they're part of a single numbered list. You'll end up with 7 steps in total (for mobile), which is more than the recommended 6, but considering this process is an important one, we can go ahead with 7. You can also merge these two steps into a single one to narrow down the number of steps to 6: 

        - (2) Select an account you want to send your crypto from.  
        - (3) Tap Send. 
        - -> (5)  Select an account you want to send your crypto from and tap Send. 

    - I'm not concerned so much about the number of steps on desktop, as desktop designs will change to align with mobile. But you can merge the same steps to narrow down the steps to a total of 7. 

    - Select an account you want to send your crypto from -> Select ~~an~~ _the_ account you want to send your crypto from. 

    - In this new step (5), explain to the reader they need to repeat this process for every account in the wallet they believe is compromised (and still have funds/NFTs). 

    - (3) Select your legacy address [..]. I would remove the _legacy_ word here; I understand your goal, but it will make the reader think, 'What's the legacy address?'

    - Switch back to your old Status profile -> Switch back to your ~~old~~ _previous_ Status profile ? Or '_compromised_ Status profile', to be more precise?
    
Thanks.

shamisem:
Thanks @jorge-campo, I revised the article using your suggestions - all of them 🙂

jorge-campo (commented):
Hello Emma – Thanks for this. 

I have several review comments, mostly about content structure, with a suggestion to add one piece of information (see **Note** below). 

Happy to discuss this on our sync call if you find that helpful. 

## Intro  
- There is no `if-someone-else-has-access-to-your-wallet` document. I'm not sure you forgot that one, or if it's included on a different branch. 

- "If there are transactions in your history […]" -> "If there are transactions in your _wallet_ history […]" ? If the user lands in your document and has no previous context, it may be hard for them to know what you're talking about. 

- "If you see the same unauthorized transactions […]" -> "If you see the same _unrecognized_ transactions […]" We don't know yet if the transactions are unauthorized; they may be valid transactions the user forgot about. 

## Create a new Wallet 

I have doubts about this H2's title. Creating a new wallet is the _action_ I need to complete to protect any remaining funds after detecting an unauthorized transaction. However, _my goal_ is not to create a wallet but to protect my funds (if that makes sense).  

Moreover, users can't really create a new wallet. They can only create a new Status Profile; both things go hand in hand. I would consider a title like "Transfer your remaining funds to a new wallet". 

After your paragraph, I would use a tip-style admonition to let the user know they can use any Ethereum-based crypto wallet to transfer the remaining funds; they don't need to create a new profile in Status if they don't want.  

"[…] create a new wallet and […]" 

- In Status, users can't directly create a new wallet. They can only create a new Status profile. 
- "Creating a new account […]" -> What about "Creating a new _Wallet_ account …" It's always useful to remind users that when we talk about accounts, we aren't talking about their Status profile. This will be a source of confusion because "account" is commonly associated with your Instagram or Telegram account; it's good to clarify the terminology whenever we can. 

- "For full security, you need to create a new recovery phrase." Yes, the user needs a new recovery phrase, but to create a new one, they need a new Status profile. Creating a new recovery phrase is not something the user can (directly) do; therefore, we need to refer to what the user can do. I think something like this works better: "You need to create a new Status profile. This action creates a new Status Wallet and recovery phrase automatically." (or similar). I would also remove the "For full security" wording and use "must" instead of "need to" (yes, we need to be assertive here).
- Link to the article where we explain what's a recovery phrase. Your article can be even more helpful for the user with more cross-links.

> **Note**: We provide lots of good information in this article, but here is the information I think is missing, and we should address it. Let's imagine for a moment I see one or more transactions in my Status Wallet that I don't recognize. I verified these using a block explorer and concluded these were unauthorized transactions. I still have funds in my Wallet accounts that I want to protect. At this time, what I need to do is to create a new Status profile (which, in turn, creates a new wallet address) and transfer my funds from the cracked wallet into the new one. 

> However, doing so has a big disadvantage. My profile information, my contacts, chats, communities, etc., are all in the previous profile (the one with the compromised wallet). If I don't use my profile much, this can be a minor inconvenience, but if I have plenty of connections and communities in that profile, this can be a major burden. And If I am a community owner, I must transfer my private's community key onto the new profile. 

> Unfortunately, if your wallet has been hacked ("hacked" is not the right term here; more on this below), you should no longer use your profile. Even when your article addresses the compromised wallet, it should mention what happens with your profile as well. Users must be aware that if someone has access to their wallet, they also have access to their profile, and therefore, they should never use that profile again. 

## Protect your Wallet in future 

- About the Structure of the checkboxes. 

    - Write a hard break between the checkbox title in bold typeface and the rest of the information. This is (somewhat) explained in the style guide (see https://write.status.im/style-conventions/#bullet-lists).

    - Your paragraphs are too long; it makes the text "less appealing" to readers who usually don't read. You need to split them into smaller chunks. You can also shorten the text; for example, you say, "you should …. are more secure". Simply say you should protect your recovery phrase and link to the corresponding article. 

- "Protect your Wallet in future" -> "Protect your Wallet in _the_ future" ?
    - "Learn more how to protect your crypto". This sentence looks broken. What about "Learn how your crypto Wallet works" or "Learn more about how to protect your crypto".
    - "Anyone who has your recovery phrase can manage your crypto." -> "Anyone _with_ your recovery phrase can _access_ your crypto _and Status profile_." 
    - "To learn more about common crypto scams..." works better as a new checkbox, making the chunks of information smaller and easier for the eyes.

- "Be careful with downloading new apps" 
    - "double-check every app you install" -> "double-check every app _and browser plug-in_ you install". 

- "Avoid using public Wi-Fi networks" 
    - Again, this is a big block of information. Please, see how you can split the information into smaller parts. 
    - "As a rule, anyone can scan phones and other devices that are using such networks." I would say, "As a rule, _consider any public network (such as airports or cafeterias hotspots) insecure_". Your original sentence gives the impression that anyone can use a WiFi scanner and hack into your phone, which is possible but not trivial. 
    - "This means that a _hacker_ …" use "malicious user" instead, as "hacker" does not necessarily have a negative connotation in a large part of the technical community who is a target for our app. See [Hacker - Wikipedia](https://en.wikipedia.org/wiki/Hacker) for an interesting discussion on this topic. 
    - "You should also avoid …" -> I think this works better as an admonition and give some variety to the topic structure.

PR #650, opened 2023-05-25
Add or remove a nickname from someone else's profile

jorge-campo:
Hello @Fabiomorais87 - I'm sending this one for your review.

Thanks.


jorge-campo:
Thanks for your review, @Fabiomorais87 – Please, see my answers to your comments inline.

> Your first paragraph introduces the idea of nicknames and what the user can gain from them. The concept of the nickname and the benefit behind it is a great one, but maybe it could put on a more conversational tone. I thought about few things to improve it a bit.

Thanks for your suggestion, Fabio. I believe the concept of 'nicknames' is well understood by our audience. Besides, this is supposed to be a quick and easy-to-do thing. Considering these, I believe it's better to move the user toward the action as soon as possible.

> In regards to this first step, when I first read the first step, I questioned if I could assign a nickname to someone I don't know or haven't added to my contacts. Since this is possible, I suggest mentioning this in the article, perhaps as an admonition, to keep channel on the step.

Good point! I included that information in the topic.

> Apart from that, do you mean "group chat" by chat? If so, I think we should use "group chat" instead of "chat".

When you say "chat", it includes DMs and group chats. I updated the procedure, just in case this is not clear.

> **Second step** - "1. Long press the person's name and select :mobile-edit: **Add nickname**."

I think I wrote this procedure before we decided to avoid the "long press" action. Thanks for the heads-up; I've updated the procedures.

> **Third step** - "1. Enter your desired nickname and tap **Save**."
> 
> For the third step, I suggest changing "Save" to "Add nickname". This is because this is what the button on the figma file is called.

I followed the dark mode workflow, which uses "Save". This is an error, and it should say "Add nickname" instead. I have updated the procedure and notified the Design team.

Screenshot 2023-07-06 at 10 38 17@2x

> I couldn't replicate this step. The Figma files that I have access to I couldn't find this option, and in the app (release 0.12) do not match the description of the step. To remove a nickname I followed the same procedure to add and then I click on an "x" that appears next to the box where I write the nickname (I'm not posting a picture because when I try to take a screenshot, it does not appear this "x" that I am mentioning). May we check together this last procedure?

I verified the procedure again on my Status desktop installation and it works as described. The only thing I realized is that right-click does the same thing as the standard click, so I removed the right-click action.

jorge-campo:
Thanks for your, review @Fabiomorais87 !

jorge-campo:
Waiting to merge this content into the new Status Help.

PR #654, opened 2023-05-31
Grant exclusive access with tokens

jorge-campo:
Using tokens, community owners in Status can configure permissions to restrict access to the community or particular channels within it. 

This article provides a step-by-step example of how to use these token-based permissions to host an event or conference. Other examples will be provided in the upcoming #653. 

These examples' goal is to provide community owners with use cases on how to use token-based permissions in Status Communities.


cheny0 (commented):
Only one comment from me:

jorge-campo:
Thanks for your review, @cheny0 ! The document has been published :smile:

PR #656, opened 2023-06-02
About changing your password

cheny0:


jorge-campo (commented):
Thank you, Chen – All is good with the content, but the structure can improve.

- About changing your password -> I would add _Status_ in this case: About changing your _Status_ password. Why? Because our app has different functions, including the Wallet. We need to inform users that there is one password for all of them, protecting their Status profile and Wallet.
- The feature availability goes in an admonition without a title (you can check this out in our style guide). Your first sentence concerns feature availability, so that should be an admonition.
- Your tip admonition is somewhat out of context. An admonition is an information that clarifies or extends other information. In this case, your information is the admonition (at least until we implement the password change functionality).

jorge-campo (changes requested):
(Please, see my review notes and the changes to include in the previous comment)

cheny0:
Thank you @jorge-campo. I have applied your comments and merged the content.

PR #659, opened 2023-06-06
Send direct messages to your contacts

cheny0:
- Create send-direct-messages-to-your-contacts.md
- Final draft


jorge-campo (changes requested):
Thanks for your article, @cheny0 !

Here you have my review comments. Just a minor change and, other that that, the article looks ready for publication.

"A direct message is a communication between you and another person". Use the DM acronym here, which we use in other parts of our documents. This would read like "A direct message _(or DM, for short)_ is a communication between you and another person."

cheny0:
> Thanks for your article, @cheny0 !
> 
> Here you have my review comments. Just a minor change and, other that that, the article looks ready for publication.
> 
> "A direct message is a communication between you and another person". Use the DM acronym here, which we use in other parts of our documents. This would read like "A direct message _(or DM, for short)_ is a communication between you and another person."

Thanks @jorge-campo! I have applied your comment and merged the content.

PR #666, opened 2023-06-12
Understand common scams

shamisem:


jorge-campo (changes requested):
Excellent content, @shamisem ! Thank you.

Here are my review comments; mostly minor things except one open question for John.

- "Status Wallet is decentralized and..." This is a good opportunity to link to the Status Wallet's quick start guide.

- Regarding links; I think you should link somewhere in your content to your other related article, "If you see transactions you don't recognize" (https://github.com/status-im/help.status.im/pull/651). If someone takes ownership of your wallet by tricking you into sharing your recovery phrase, you'll surely see transactions you don't recognize. Since you are already linking that article to this one, it would be appropriate to link both of them reciprocally.

- "There's only one official Status website, [status.im][status-im], and we don't have any others." Please, ask John on the chat what will be the final website. I believe it will be `status.app`, but I don't know what will happen with `status.im` (I guess it will be redirected). If he doesn't answer, this is a good question for next week's meeting with him.

- "Scammers can create fake apps and extensions that steal..." I would say "and _browser_ extensions that steal..."

- "You can also try googling its name..." I prefer we use something more generic and avoid the pronoun, like "You can also search the company's or extension's name online..."

- "Such tokens are not visible in your Status Wallet." Are we sure about this? Honest, non-rhetorical question.

- "To prevent this, only use trusted crypto swap services." The precedent is somehow unclear with _this_. What do you think about "To prevent this _scam_, only use trusted..."

shamisem:
Hey @jorge-campo, thanks again for your review!

All done. As for `"Such tokens are not visible in your Status Wallet." Are we sure about this? Honest, non-rhetorical question.`, Pedro thinks they won't be visible too. I'll double-check with John on our next meeting and edit this sentence if it turns out to be wrong.

PR #668, opened 2023-06-13
570 Customise mentions and other messages

Fabiomorais87:
Hi @jorge-campo,

Fabio


jorge-campo (changes requested):
Hi @Fabiomorais87 – Here are my review comments.

I'm aware we have reviewed this content together, but on a second look, it's confusing. I believe the topic's title and the tables make it hard to understand the procedure's goal. Allow me to explain here in my review comments.

Intro
---
Let's change your intro so it looks more 'coherent'. The "Know your notifications options" H2 looks unnecessary, and the title of the columns is confusing. For example, the first column's items are not really "options" but more "notifications" (if that makes sense; let me know if it doesn't).

Let's go step by step.

1. The way you start with the first sentence can make the user think these are the only things they can customize in Status. In reality, they can customize these things _for any type of communication_. In other words, the answer you should address in this topic is not what are the notification options but how I can apply these options to the different types of messages in Status. This bit of information is missing. 
2. "For each one of them..." In this intro, You're referring to a procedure later in the text. There is a big distance between this sentence and the procedure. This makes the document hard to understand because you ask them to follow a procedure the can't see yet. Does it make sense?
3. Address what the topic is about. What is this about? It is about customizing notifications for mentions and all other messages. Your content seems focused on the UI elements (the different types of notifications) and not how to customize them. (This is similar to what I expressed in point (1)).

I propose you the following:

- Change the title of the topic to "Customize mentions and other messages notifications"
- Rewrite the first sentence. Something like, "When using the Status app, you can customize how you receive notifications in direct messages (DMs), group chats or communities. For example, you may want to receive notifications for mentions on group chats but not for communities, or you may want to receive DM notifications only in the [Status notification center][link-to-article-where-you-talk-about-the-notification-center]."
- Remove the second sentence ("For each one of them...").
- Add a new paragraph where you discuss exceptions. For example, "With the Status app, you can further customize the notifications you receive with [exceptions][link-to-your-exceptions-topic]. For example, you can use exceptions to 

Then, introduce the table. Something like "Here are the notifications you can customize:"
- Typo: "Personal metions" -> "Personal _mentions_"
- Rename the first column to "Notifications".
- Remove the last column. You don't need the column if all the options are the same for a column. You can mention this information elsewhere in the intro (only if you think this is relevant).
- Remove the last row, "Same settings for all message types". Even when this option is next to the others in the UI, it has nothing to do with the others. The table is about the different types of notifications; the "Same settings for all message types" is not a type of notification. You're mixing two different concepts in one table.
- Remove the "Note" admonition; this is already explained in the new intro. Use here the last admonition in your document, "You can't customize...", but make the sentence positive: "Currently, you can only customize your channel notifications on the Status desktop app."
- Don't need to overexplain the options in the description column:
    - Receive messages that specifically mentions you" -> Messages that specifically _mention_ you." (_mention_, singular).
    - "Receive notification for announcements ..." -> "Messages directed to the entire community."
    - "Receive notification for messages..." -> "Any message that is not a personal or global mention".
    - (we remove the last row)

"Customize mentions and all other messages" procedure
---
- Change the H2 title to "Customize mentions and all other messages _notifications_".
- Change the first sentence preceding the procedure. What you really choose is _the way_ you receive (or don't receive) notifications. Suggestion: "You can customize how you receive notifications for personal and global mentions and for all other messages."
At the end of this procedure, use here a tip-style admonition where you explain the purpose of the "Same settings for all message types". For example, "Tip: You can configure the same notification setting for DMs, group chats or communities using the `Same settings for all message types` option."
- The first step is too long. Suggestion: "From the tab bar, tap  Communities to customize your community or channel notifications, or tap  Messages to customize your group chat or DM notifications.
- Step (2) mentions "Long press the community..."; can we do this without the long-press action? (due to accessibility).
- Step (5) "Choose your notification preference." -> "Choose your _preferred_ notification _setting_:"
- Table on this procedure.
    - Change "Preference" column to "Setting".
    - "Turn on this if you want ..." -> In this case, the style guide rule for phrasal verbs doesn't work. Use "Turn this on if you ..." instead.
    - "Turn on this option to receive subtle notifications..." It may be difficult to understand what we mean by subtle. Just use, "Turn this option on to receive the notifications in the Status notification center. You need to unlock your phone to see the notifications."
    - "Turn this off if you don't want to receive notifications". Isn't it "Turn this _on_ ..."?

"How to disable Status on-screen notifications" procedure
---
- Change the H2 to "How to disable on-screen notifications".
- Start explaining what are "on-screen notifications". For example, "On-screen notifications appear as banners on your device's screen when you're actively using it or when the screen is locked."
- You start with "Here is an example ... " but this is not an example; it's a procedure the user can apply to any message type. Simply say, after the introduction to what on-screen notifications are, "You can disable on-screen notifications for new messages within the Status app. This may be useful..." (note: I add "within the Status app" because you can achieve the same result using the notification settings of your device's OS).
- Step 1: Same as the previous procedure; it's too long.
- Step 2: Same as the previous procedure; see if we can reach the same point without "long-press".

If this is confusing and you prefer to discuss all these together, just let me know.

Thanks.

Fabiomorais87:
Suppressed by https://github.com/status-im/status-website/pull/850

PR #682, opened 2023-07-03
640 Reset your notification settings

Fabiomorais87:
Hi @jorge-campo,

Fabio


jorge-campo (changes requested):
Thanks for this content, @Fabiomorais87 – Here are just some minor changes I propose you to include:

- Intro. We can simplify the wording a bit. I used WordTune for this. Same number of words, but the wording is easier: 'If you set custom notification settings before, but you don't receive the notifications you expect anymore, you can reset your settings. Doing so removes any exceptions, and the notification settings return to their default values.'
- Procedure – You're missing the last step, where the user confirms they want to reset the notifications. Whenever you do these types of 'danger' actions (reset, delete, clear, etc.), it's almost certain you need to confirm the action. See [this Figma file](https://www.figma.com/file/JlpPhJp0SMnEyBQy4nOZHo/Settings-for-Mobile?type=design&node-id=4312-388322&mode=design&t=SdtLYcdM3ZTw2cau-4).

jorge-campo:
Hi @Fabiomorais87 , what's the status of this? Could you let me know if you implemented the review comments? Thanks.

Fabiomorais87:
This article has already been reviewed by campo and the content is good for publication. I already sent it to .mdx syntax review here https://github.com/status-im/status-website/pull/586. Waiting for the pr to be approved and then merge the content.

PR #686, opened 2023-07-05
Change a token-gated channel to open

### LINKED ISSUE #638: Change a token-gated channel to open

jorge-campo:
Hello @Fabiomorais87 – Here is a new article for your review.

Thanks!

---

- 638 change terminology 'private' -> 'token-gated'
- 638 first draft -- ready for review

### DIFF
--- a/config/en/mkdocs.yml
+++ b/config/en/mkdocs.yml
@@ -254,7 +254,7 @@ nav:
     - Administer channels:
       - Create a channel: status-communities/create-a-channel.md
       - Customize your channel: status-communities/customize-your-channel.md
-      - Change a private channel to public: status-communities/change-a-private-channel-to-public.md
+      - Change a token-gated channel to open: status-communities/change-a-token-gated-channel-to-open.md
       - Set up channel permissions: status-communities/set-up-channel-permissions.md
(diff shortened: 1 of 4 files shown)


jorge-campo:
Thanks, @Fabiomorais87 – I'm adding my answer to your comments review here and closing this PR. If you have additional comments, you can add them to the new PR in https://github.com/status-im/status-website/pull/552

> I think it would be interesting to use the words **join** and **post**.

We no longer use the term 'join' for channels, only for communities. The concept is simple: you can click/tap on any channel you have access to based on the channel configuration and the assets you own. Everything happens transparently for the user.


I suggest replacing the words symbol and icon with the following: Token-gated channels show a :mobile-locked-lock: lock icon next to the channel name.

I changed "icon" but not "symbol". We should refer to those as symbols.


> If my suggestuin mekes sense you should choose just one word.

Makes sense. I've updated the paragraph following your suggestion 👍 


> I could not replicate the procedure on the desktop app (0.13 release). It may be because the desktop app has not been updated yet with the Figma information.

I have reproduced the scenario in the topic, and the procedure works as described in 0.14.1

PR #688, opened 2023-07-05
Import an account 

earakel:
Hey Jorge, sending this for your review!

The article is missing the Mobile section for Keycard, as the designs are still in progress. I still think we can publish this since the desktop version is ready.

### DIFF
--- a/config/en/mkdocs.yml
+++ b/config/en/mkdocs.yml
@@ -286,6 +286,7 @@ nav:
     - status-wallet/index.md
     - Getting started:
       - Status Wallet FAQ: status-wallet/status-wallet-faq.md
+      - Import an account into your Status Wallet: status-wallet/import-an-account.md
     - Send and receive crypto:
       - Receive crypto: status-wallet/receive-crypto.md
       - Send crypto to your Status Wallet: status-wallet/send-crypto-to-your-status-wallet.md
--- a/docs/en/status-wallet/about-custom-derivation-paths.md
+++ b/docs/en/status-wallet/about-custom-derivation-paths.md
@@ -1,0 +1,16 @@
+---
+id: 687
+revision: 0
+language: en
+hide:
+  - navigation
+---
+
+# About custom derivation paths
+
+ :octicons-tools-24: In progress
+
+!!! note ""
+     We're working on this content.
+
+--8<-- "includes/urls-en.txt"
--- a/docs/en/status-wallet/import-an-account.md
+++ b/docs/en/status-wallet/import-an-account.md
@@ -8,9 +8,75 @@ hide:
 
 # Import an account into your Status Wallet
 
- :octicons-tools-24: In progress
+If you already have crypto accounts, you can import them into your Status Wallet. To do so, you can use your recovery phrase, private key or Keycard.
 
-!!! note ""
-     We're working on this content.
+!!! note
+     Status Wallet only supports Ethereum and Ethereum-based chains.
+
+## Import your recovery phrase or private key
+
+To use your existing account with Status Wallet, import either your recovery phrase or private key. Both options are valid and retrieve the same account for you.
+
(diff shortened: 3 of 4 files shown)


jorge-campo:
> The article is missing the Mobile section for Keycard

Checking @shamisem; reviewing it now – To keep track of what's missing, please, create a new 'update' issue in your GitHub project board with a note about what's missing now and needs to be added in the future update. Thanks.

jorge-campo (changes requested):
Hi Emma - Great content! However, you're missing plenty of icons and formatting in the procedures. Please, don't overlook this.

Here are my review comments.

- Your Markdown filename and article name are different: Import an account into your Status Wallet / import-an-account.md. Please, update the Markdown filename, URLs, mkdocs.yml, etc., so they use the same name.

### Document layout

I would structure things differently. As you use the H3s, the 'Import an account' H2 doesn't provide added value. Think about the tasks before writing them (task analysis). The tasks here are:

    a. Import with a recovery phrase
    b. Import with a private key

And under 'Importing with a private key', there are two methods: the recovery phrase and Keycard.

Once you have this in mind, finding the right headers is easier. Your headers would be:

```
H2 Import an account using your recovery phrase
H2 Import an account using your private key
H3 Import your private key manually
H3 Import your private key using Keycard
```

### Intro 

- 'If you already have crypto accounts...' -> Suggestion: 'If you already have accounts in another crypto wallet...'

- Admonitions. We should use at least one warning-style admonition in every article where we ask the user to do things with their private key or recovery phrase. What would this look like?

```
H1 Import an account into your Status Wallet
H2 Import an account using your recovery phrase
-> Admonition about the risk of disclosing your recovery phrase.

H2 Import an account using your private key
-> Admonition about the risk of disclosing your private key

H3 Import your private key manually
H3 Import your private key using Keycard
```

- Steps. You have too many steps on the mobile procedures. You need to combine them and narrow things down to 6 or 7 (maximum). If you can't do so, you need to split the procedure into two bigger steps with individually numbered lists within them.
- You're missing the icons in all steps (!) Please, check the icons you need from the Figma files (desktop and mobile) and add them to your PR. We'll change the icon names once we move to the new Status website, but we need the icons first to change properly.

### H2 Import an account using your recovery phrase.

Tell the user, in one sentence, what's a recovery phrase. For example, 'A recovery key (or seed phrase) consists of several randomly generated words created when you first set up a wallet.'

- Mobile
    - In step 3 and step 4, 'Origin' and 'Import New' are bold typeface options. See this example:

    Screenshot 2023-07-12 at 10 56 58@2x

    - Step 5 should read 'Select *Import using recovery phrase*', as depicted on the Figma file.
    - You can combine steps 3 and 4 AND/OR 5 and 6.

- Desktop
    - 'Click your profile picture under Origin.' 'Origin' should be in bold typeface.
    - 'Under Import new, select :desktop-key: ...' 'Import new' should be in bold typeface.

- Include an admonition about the risks of handling your recovery phrase improperly. We have this admonition in one document:

> Keep your recovery phrase safe and never share it with anyone. Status will never ask you for this information, and anyone who does is trying to steal your profile and wallet funds

### H3 Import an account using your private key

Tell the user in one sentence what's a private key. For example, "A private key is a string of letters and numbers that allows you to access and manage your crypto funds."

- See the same formatting recommendations from the previous procedure. You also need to add icons in the majority of steps.
- Reduce the number of steps using the same approach discussed in the previous procedure.

### H2 Import a private key from your keycard

Users may not know what's a Keycard. Start the procedure with a simple sentence explaining what's a Keycard and where to find more information (link to the corresponding 'About' article). 

- Steps 3, 4, 5: 'Origin', 'Import new', and 'Settings' all have bold typeface.
- 'Click Name account (or Name accounts, if you have multiple accounts on this Keycard)' You're missing the final dot (.)

Thanks.

PR #692, opened 2023-07-12
About voting to change community visibility

cheny0:
- Ready for review
- Update the title


cheny0:
Hi Patryk @osmaczko, this document introduces the voting mechanism. I'm aware that the content about "Vote to feature a community" is not final, we can ignore that part for now. Thanks!

docs/en/status-communities/about-voting-to-change-community-visibility.md
```diff
@@ -0,0 +1,67 @@
...
+  - navigation
+---
+
+# About voting to change community visibility
+
+All SNT holders can engage in [voting to include or exclude a community][vote-to make-a-community-private-or-public] from the Communities Discover page, and [voting to feature a community][vote-to-feature-a-community]. 
+
+Votes happen on the [community directory curation dApp][curate.status.app]. This is an independent, decentralized application deployed on the Optimism network and is separate from the Status app. Voting outcomes are solely determined by the voters, and Status has no control over these results.
+
+When a community is created, it's private by default, which is ideal for small, exclusive communities. However, if the community targets a broad audience, you might prefer to make it public.
```
osmaczko, line 15: Technically, the community is never private. An unencrypted community description is broadcasted on the Waku network as soon as it's created, implying that anyone can read it.

Perhaps instead of saying "it's private by default," we could state that "it is not listed in the global directory."

docs/en/status-communities/about-voting-to-change-community-visibility.md
```diff
@@ -0,0 +1,67 @@
...
+
+## The basics
+
+- Every SNT holder, not just community members or Status users, can participate in all voting steps.
+- Your voting power is determined by how many SNT tokens you stake.
```
osmaczko, line 20:
```suggestion
- Your voting power is determined by how many SNT tokens you hold.
```

docs/en/status-communities/about-voting-to-change-community-visibility.md
```diff
@@ -0,0 +1,67 @@
...
+- You don't spend the tokens to vote, but you must hold the tokens during the validation and finalization periods, or your vote is discarded.
+- You need to pay a [gas fee][understand-network-fees] in ETH for initiating, validating and finalizing a vote. The amount of gas fee depends on the traffic of Optimism.
```
osmaczko, line 22: Maybe it is worth to mention that most users don't need to pay gas, the voting itself is free.

docs/en/status-communities/about-voting-to-change-community-visibility.md
```diff
@@ -0,0 +1,67 @@
...
+    Status does not charge you any fees for voting.
+
+## Voting to make a community public or private
+<!--
+(Not implemented yet)
+When initiating a vote, your minimum stake is ~200k SNT. However, for every unsuccessful voting attempt, the minimum SNT required to start a new vote doubles.
+-->
+The voting period lasts for 14 days.
+
+If a single vote exceeding 2,000,000 SNT supports the community's removal, the remaining voting period reduces to 24 hours. However, this can be reversed if a single vote of over 2,000,000 SNT opposes the removal.
```
osmaczko, line 35: This is also not implemented yet.

docs/en/status-communities/about-voting-to-change-community-visibility.md
cheny0: Using "stake" rather than "hold" because users can choose the number of SNT tokens for voting, it's not necessarily the amount they hold.

docs/en/status-communities/about-voting-to-change-community-visibility.md
cheny0: We name "communities that are not listed in the global directory" as "private communities". I'll rephrase this sentence to avoid confusion.

docs/en/status-communities/about-voting-to-change-community-visibility.md
cheny0: I have applied your comments. Thanks!

jorge-campo (changes requested):
Thanks for your work on this not-so-easy article, @cheny0!

Here are some suggestions to improve your content:

## Title

Your title is missing an article. I verified this point with Bing Chat, and it's better to use 'About voting to change *the* community visibiliy'. Don't forget to rename the Markdown file as well.

## Intro

I suggest some changes in the first paragraph. Avoid using filler words like 'engage' and passive forms. 

'All SNT holders can vote whether to include, exclude, or feature a Status Community on the Communities Discover page within the Status app. Only voters decide the outcome of the vote, and Status has no control over the voting results.'

After this first paragraph, we need a screenshot with the "Discover" screen. Otherwise, it might be challenging for the reader to understand what we're discussing. Alternatively, you can add the screenshot as the header of the article. It doesn't matter if it's the same header you used in another article, as long as it's clear what we're talking about.

## The basics

- You don't spend the tokens to vote, but you must hold the tokens during the validation and finalization periods, or your vote is discarded.
    - 'Discarded' has a somewhat negative connotation. Whenever you mention something negative, try to make it sound positive. The only exception to this 'rule' is when you discuss security-related things or things that can result in loss of funds.
    - An SNT is a token, correct, but it's a particular type of token. It's an asset.
    - I suggest something like, 'You don't spend or lose control over the assets when you vote, but you must hold the assets during the validation and finalization periods for your vote to count.

- 'The amount of gas fee depends on the traffic of Optimism' -> 'The amount of gas fee depends on the traffic of the Optimism _network_' and link 'Optimism network' to the 'Layer-2 scaling solutions for Ethereum' article.

## Voting to make a community public or private

- 'When a community is created, it's a private community by default'. In 'private community', link to the different types of Status Communities article.

- There is something I don't understand, and that may be the case for the reader as well. You say, 'If no one verifies the vote, only the initial vote is counted', and then you say, 'After verification, the vote must be finalized.'
    - Your first sentence seems to imply that it's possible that no one verifies the vote (by 'vote' I assume you mean all the votes?), but then you say that after verification, the vote must be finalized. So, it isn't clear to me if verification is mandatory; or what happens if no one verifies the vote; is the verification automatically closed somehow? (or something else). Let me know if this makes sense or if you prefer to discuss this on a call.

## Voting to feature a community

- In the same way, you explain in the previous section what happens with private a public communities, you need to explain here what it means to 'feature' a community. Actually, you have that sentence in the second paragraph, 'The top five communities in the number of SNT tokens appear in the Featured section, replacing the communities that were featured previously.'

- ' If no one verifies the vote, only the first vote of this round is counted.' I have the same problem with this sentence described in the previous section. Is verification 'mandatory'? What happens if no one verifies the vote?

- 'The Feature this community button next to the name of this community is greyed out' -> 'The Feature this community button next to the name of this community is greyed out _and disabled_'.

## About finalizing a vote

- 'Otherwise, the vote remains pending.' -> 'Otherwise, the vote remains pending _and no more votes can occur._'

- It's worth highlighting again, using a note-style admonition, the fact that 'The top five communities in the number of SNT tokens appear in the Featured section'. Users should not assume that just because the vote is favorable to featuring the community, it will be featured.

## Note

On each one of your H2s, I miss a link to the article where you explain how to do these things. You have two procedure-type articles that you can link at the end of every H2 section. Even when this may look repetitive, it's a good repetition because it helps the reader understand the real process.

cheny0:
Hi @jorge-campo, I have edited the document based on your comments and our discussion. Thanks!

PR #695, opened 2023-07-17
290 Understand token requirements in channels

Fabiomorais87:
Hi @campo, the following is the first draft of the article "Understand token requirements in channels". I would appreciate it if you could review it. This is just for checking if the content is correct. 

I don´t know why I wasn't able to link to the issue, that's the reason why is not linked.

### DIFF
--- a/docs/en/status-communities/understand-token-requirements-in-channels.md
+++ b/docs/en/status-communities/understand-token-requirements-in-channels.md
@@ -8,9 +8,24 @@ hide:
 
 # Understand token requirements in channels
 
- :octicons-tools-24:
+In Status, there are two kinds of channels: open and token-gated. If you want to [join a token-gated channel][join-a-channel], you need to meet the token requirements. To maintain access to a token-gated channel, hold the necessary tokens at all times.
 
-!!! note ""
-     We're working on this content.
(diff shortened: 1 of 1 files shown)


jorge-campo (changes requested):
Thanks, @Fabiomorais87 – I'm afraid we need to rewrite this content.

The biggest problem with the article is that you're overlooking the added complexity of configuring tokens in channels. Your content copies 'Understand token requirements in communities' but misses the mark in explaining the different token requirements in channels: view-only AND view and post.

This is key in understanding token requirements in channels. Tokens are not (as it occurs in communities) a simple way of either accessing or not accessing a channel. Users can:

a. Don't meet any of the channel requirements.
b. Meet only the view-only requirements, so they can't post content.
c. Meet view and post requirements; therefore, they can post content.

You also need to mention that meeting the view and post requirements allows users to read and post independently of meeting the view-only requirements.

Because of this, your example in the second paragraph about the 15 SNT falls short of helping readers to understand the nuances of token-gated channels. In the same way, your two H2s are not enough either, as there is no binary "you meet / you don't meet" the requirements in channels.

Please, do a task analysis of this content and try to anticipate the reader's questions and tasks when working with token-gated channels. Then, write the H2s you need to address these questions. Don't just copy and paste the 'Communities' content, please. Think about the content.

Plan what screenshots you want to include; I believe all the screenshots you need for this content are already in place, and you can use them (I'm talking about the fictional 'Acme' community in Figma).

Send me a new pull request when you have the content ready.

Thanks.

PR #693, opened 2023-07-12
Vote to make a community private or public

cheny0:
- First draft
- Final draft
- Minor fix

PR #188, opened 2022-07-13
use requirements.txt in build docs

jakubgs:
You don't want to maintain package versions in two separate files.

You are asking for divergence between actual dependencies and the install instructions.

PR #698, opened 2023-07-21
Vote to feature a Status Community

cheny0:


docs/en/status-communities/vote-to-feature-a-status-community.md
```diff
@@ -0,0 +1,78 @@
...
+## Step 2: Cast your vote {: #cast-the-vote}
+
+The voting period lasts for 5 days. Voters don't need to pay any gas fee.
+
+1. From the top navigation, select **Directory** to view all public communities.
+1. Find the community you want to vote for and select **Feature this community!**.
+1. Choose the amount of SNT to stake and select **Confirm vote to feature communtiy**.
+1. Confirm the transaction and pay the gas fee from your wallet.
+
+## Step 3: Verify the vote {: #verify-the-vote}
```
osmaczko, line 59: We should emphasize this step is optional, but the more peers participate, the better it is for vote integrity.

cheny0: I have applied your comments. Thanks!

jorge-campo (changes requested):
Thanks, @cheny0 – Overall, great article.

I have included here some notes that I believe can improve it. Let me know if something is unclear.

## Intro

- 'vote for it to be featured.' -> '... vote for it to be featured in the Status app.'

- After your second paragraph, having a screenshot with the mobile app screen is helpful, so users can refer to the place in the app you're talking about. This screenshot is ready here: https://www.figma.com/file/PmChAIrzndYam0aNWkcrgz/Documentation-for-Mobile?type=design&node-id=845-245406&mode=design&t=fZ6Vu3ByBogjVs4u-0
    - Please, ask Francesca to modify the screenshot so it only shows the upper half of the mobile screen, with one or two fake communities listed under the 'Join' tab.

## Vote to feature a Status Community

- This header and its content work better in bullet points. This would be a 'The basics' section.

- '...but you must hold the tokens during validation and finalization periods.' At this point, the reader doesn't know yet what the 'validation' and 'finalization' periods are. I would change your sentence to say, ' ... you must hold the tokens you stake until the voting process finishes.' (or similar)

- 'A gas fee in ETH is required ...' There is no reason here to use a passive form: 'You should pay a gas fee in ETH for initiating, ...'

- Place your first admonition here. This is where the information is relevant because here is where you're talking about fees. Your first admonition, as it is now, looks out of place.

- I would make your second admonition a tip. This information is just nice to have, and a tip adds more variation to your admonitions.

- After the colon, remember to use lowercase, except for product names. See https://write.status.im/style-conventions/?h=ca#capitalization for reference.

The items in your checklist don't have a final period because they're not complete sentences. See https://write.status.im/style-conventions/?h=ca#bullet-lists for reference.

## Step 1

The 'Community Directory Curation dApp' link is at the beginning of your article. We can't include the link in the procedural steps, but I think it's a good idea to include the link again in the intro paragraph. This is where the user will look for the link. Maybe, you can say, 'Make sure you can open the Community Directory Curation dApp [link here], and your wallet contains ...' It's just an idea.

- 'switch the network to Optimism' You need to explain how to do this. Focus on the Status Wallet how-to. In Status Wallet, you don't 'switch' to one particular network; you check or uncheck (select) a network from your list of networks. This is what you need to explain. You can find the Figma file for mobile here, https://www.figma.com/file/HKncH4wnDwZDAhc4AryK8Y/Wallet-for-Mobile?type=design&node-id=1-46573&mode=design&t=337K49o5iTGL65fY-4, and you can use you Status desktop app to check the process in desktop (it's almost the same).
- In this case, I wouldn't make a procedure for mobile and another for desktop. Add in the same sentence how to check the Optimism network for mobile and for desktop.

- About the wallet compatibility. So far, we've checked the MetaMask wallet and the Status one. This doesn't mean that other wallets don't work on this dApp. For this reason, I think it's better if we explain in the first sentence that the process is for Status Wallet, and then, on your admonition, you can say that other crypto wallets may work in addition to Status. As it is right now, it seems that we only support two wallets for the dApp (Status and Metamask), but this is simply a dApp; there is nothing that prevents other wallets from working with it.

## Step 2

- I believe this works better when you change the order of your sentences, firs sentence being 'A round of vote starts ...'
- 'After a previous round of vote...' This is confusing because you're assuming there is a previous round, and maybe, there is no previous round. What you want to express is the fact that _if there was a previous_ round, you need to wait at least one week to initiate another one. Correct me if I'm wrong and not understanding this paragraph correctly.

## Step 3

- '... lasts for 5 days.' '5' should read 'five'.
- Voters don't need to pay any gas fee.' Don't _need to pay_ or _don't pay_ :) 'need' introduces unnecessary ambiguity to your sentence precisely when the reader needs assertiveness.
- You don't need the dot at the end of the '... select Feature this community!'
- 'From your wallet, sign the vote.' Do we know what UI element users must click on Status Wallet for this? If we know it, you can add to your sentence, 'In Status Wallet, click XXX to sign the vote.'

## Step 4

- '... a 2-day verification...' '2' should read 'two'.
- 'From your wallet, confirm and pay the gas fee.' Again, if we can explain the UI to click in Status Wallet, let's include that.

## Step 5

- 'After the 2-day ...' '2' should read 'two'.
- 'Otherwise, it stays pending and no one can initiate a new round' I think we need to clarify what 'it' is. I would rephrase this sentence, 'Otherwise, the current voting period can't conclude and no one can initiate a new round.'
- Your last admonition didn't sound right, so I asked ChatGPT 😄  It suggested this one instead, 'After the finalization process, the top five communities with the highest number of SNT tokens will be showcased in the Featured section.'

## More about the steps

- Your steps start with 'From the top navigation, ...' However, these steps don't occur one after the other _in the same day_. Therefore, we need to say from the top navigation of what thing, because readers can arrive at your document and move straight to step 3 (for example) without reading anything before that point. With the text now, users may think you're talking about the Status app. 
- Gas fees are important to users; they don't want to spend money if they can avoid that. Therefore, I would include a first sentence for each step: "This step requires paying gas fees" or "This step doesn't require paying gas fees." This sentence goes before anything else. What do you think?

jorge-campo:
> * 'switch the network to Optimism' You need to explain how to do this.

Sorry, I just saw your comment in #693 https://github.com/status-im/help.status.im/pull/693#issuecomment-1665050437.

In this case please, create an 'issue update' for this doc in the new status-website repository so you don't forget to include that in the future. Thanks!

cheny0:
Hi @jorge-campo,

About step 2, by "a round of vote" I mean the voting process containing all five steps. This "Weekly featured vote" is not really "weekly" because if no one votes, a new round won't be started. And if no one finalize a round of vote a new round can't be started. It is weekly only when each round can be finalized immediately after the verification period, and a new round starts immediately after the previous round is finalized. 

Unlike voting to add or remove a community, here in a round of vote users can vote for whichever community that exists in the directory. (I'll add this information to the intro part) So I don't think it's necessary to tell users to wait a week to start a new round.

And I have opened an issue for adding all Status wallet related update (#188 in the status-website repo).

Other comments are implemented. Thank you!

jorge-campo:
> Hi @jorge-campo,
> 
> About step 2, by "a round of vote" I mean the voting process containing all five steps. This "Weekly featured vote" is not really "weekly" because if no one votes, a new round won't be started. And if no one finalize a round of vote a new round can't be started. It is weekly only when each round can be finalized immediately after the verification period, and a new round starts immediately after the previous round is finalized.
> 
> Unlike voting to add or remove a community, here in a round of vote users can vote for whichever community that exists in the directory. (I'll add this information to the intro part) So I don't think it's necessary to tell users to wait a week to start a new round.
> 
> And I have opened an issue for adding all Status wallet related update (#188 in the status-website repo).
> 
> Other comments are implemented. Thank you!

Awesome, @cheny0 – Thanks for the detailed explanation.

Issue #122, opened 2022-06-15
Use DApps with the Status Browser


jorge-campo:
Describe how to connect to decentralized apps or DApps using the Status Web3 Browser.

Use the content in the [Status FAQs](https://status.im/docs/FAQs.html) as the starting point:

- What is DAp.ps
- How do I vote for DApps
- What are DApp permissions

jorge-campo:
Not a priority now. Most probably available in 2023 Q3.

Fabiomorais87:
This is not available, so I am closing the issue.

jorge-campo:
Superseded by https://github.com/status-im/status-website/issues/1182

PR #699, opened 2023-07-24
About the Ethereum blockchain

jorge-campo:
Hello @Fabiomorais87 – Here is a new article for your review. Thanks!

> Note: I can't link the issue for some unknown reason.

### DIFF
--- a/docs/en/getting-started/about-the-ethereum-blockchain.md
+++ b/docs/en/getting-started/about-the-ethereum-blockchain.md
@@ -17,16 +17,21 @@ These digital transactions serve various purposes, such as buying or sending tok
 
 ## What's Ethereum?
 
-If the blockchain is the digital ledger, think of Ethereum as the computer running this ledger. The Ethereum blockchain is a network consisting of thousands of computers. Anybody is free to join and contribute to the network, but nobody owns the network. This is in contrast to corporations that run their own centralized servers, control of the network and set the rules.
+If the blockchain is the digital ledger, think of Ethereum as the computer running this ledger. The Ethereum blockchain is a network consisting of thousands of computers. Anybody is free to join and contribute to the network, but nobody owns the network. This is in contrast to corporations that run their own centralized servers, control the network and set the rules.
 
 !!! note
     Contributors to the decentralized Ethereum network receive compensation in Ether (ETH), Ethereum's native cryptocurrency.
 
-When we talk about Ethereum, we often refer to the main Ethereum blockchain or mainnet. On mainnet, Ether has a value, and transactions have an actual cost. Besides the mainnet, there are other test Ethereum networks or testnets, where Ether holds no real value.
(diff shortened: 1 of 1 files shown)


Fabiomorais87:
Hi @jorge-campo , here are my review suggestions: 

First, I would like to say that I liked the article. This article "looks simple", but in reality, it is a hard one. I say this because you needed to explain different topics in a simple way to the user. The whole time that I was reading your article, I had in mind a regular user that does not understand much about technology (like my dad). And we this in mind I can say that the article achieves its purpose. Still, I have minor suggestions that if it makes sense you could add to the article. 

## What's Ethereum?

_"When we talk about Ethereum, we often refer to the main Ethereum blockchain or mainnet. On mainnet, Ether has a value, and transactions have an actual cost. Besides the mainnet, there are other test Ethereum networks or testnets, where Ether holds no real value."_ 

I understand your point here, but in my opinion, I would take this whole paragraph out. With the regular user in mind, this paragraph would confuse the user more than aggregate more information.

Why do I think like this? The regular user does not know what is a mainnet or a testnet. This information here without more context (which clearly could not be added here) would confuse the regular user. The more advanced user already knows that.

## The role of Ethereum in the Status app

In my opinion, this intro could be expanded and linked to upcoming H3s. How about listing the following H3s in this intro? I am saying this because the H2 is the role of Ethereum in the Status app, the intro says that you can benefit from Ethereum in the app, but my following questions are "How?" and "Why should I use this app?". So in this way, in a concise way I would link this H2 to the upcoming H3s and in this way I would answer the "how" and "why" questions. Makes sense what I am saying?

### Status crypto wallet

_"You can send, receive and store tokens securely, including Ether and ERC-20 tokens,"_

I would add a cross-link to "ERC-20 tokens" to the article named as "Undestand about ERC-20 tokens".
 (https://github.com/status-im/status-website/issues/380) 

### Running dApps on Ethereum

I would add a crooss-link to an article named as "Use DApps with the Status Browser". (https://github.com/status-im/help.status.im/issues/122)

## General suggestion 

The article would benefit greatly if you added an image (header or in the introduction) that describes a blockchain. You may reuse the image that is in the article titled "About Status messages". 

Screenshot_9

jorge-campo:
Hey @Fabiomorais87 – Thanks for your review.

Here is what I have done:

- I have expanded the information about mainnet vs testnets. Why not remove this information altogether? Because there are a few screens in the Wallet and Settings section where we mention these two terms. It will be helpful for users to understand this a bit, if only for this reason. But I see your point and have discussed the section a bit further.
- I added the link to the ERC-20 tokens you suggested but not the one to the 'Use DApps with the Status Browser' article. The project needs to clarify some open questions with the web3 browser first.
- As per the illustration, I was never happy with that one 😅. I'd prefer Design creates the illustrations and headers for us using the Status brand style and guidelines.

jorge-campo:
Article is ready to merge in the status-website repo.

PR #700, opened 2023-07-25
246 Find people on Status

Fabiomorais87:
The following is the first draft of the article "Find people on Status". I would appreciate it if you could review it. This is just for checking if the content is correct.

I don´t know why I wasn't able to link to the issue, that's the reason why is not linked.


Fabiomorais87:
I am closing this PR ("draft"). @cheny0 is taking care of this article now.

